PR #24, opened 2021-12-24
Implement `pbutton`

giladt:
Implement `pbutton` instead of the old `d2d-button` element.
Introducing `full-width`option.
Apply necessary styling fixes (affecting the disclaimer page).

image


dkent600 (approved):
OK, you're not actually implementing the pbutton, you are using it.

Question I wanted to discuss:  Would it be a better design , instead of type="secondary", to have separate button elements for each type, so `pbutton-secondary`?

I'm not stuck on doing that, just wanted to discuss.

PR #27, opened 2021-12-26
Implement css-reset

giladt:
Implemented CSS-Reset package: `the-new-css-reset`.
https://www.npmjs.com/package/the-new-css-reset (v1.4.4)

> A reset stylesheet (or CSS reset) is a collection of CSS rules used to clear the browser's default formatting of HTML elements, removing potential inconsistencies between different browsers.

This eliminates the need to use only `div`s in order to get expected behavior from every HTML element and allows the use of semantic HTML which can better comply with screen readers, and other accessibility tools as well as better SEO results.


dkent600 (changes requested):
Can you pls explain in the PR description why we need this?  Why can't we set it ourselves?  Note that pbutton.scss already does this.  I think it should at least be explicit in all of the design system stuff.

PR #28, opened 2021-12-26
Added `pbutton` hover animation

giladt:
Result:
pbutton animation 

button-animation

button-is-loading

button-is-loading-toggle

### DIFF
--- a/src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
+++ b/src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
@@ -29,6 +29,12 @@
   transform: translateY(0);
   user-select: none;
   color: $White;
+  display: flex;
+  align-items: center;
+  gap: 10px;
+  &>.is-loading {
+    animation: spin 1.75s infinite linear;
+  }
 
   &:focus {
     box-shadow: 0 0 10px $Shade06;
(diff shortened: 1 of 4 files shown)


src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
```diff
@@ -3,7 +3,8 @@
 @mixin standardBackground {
   border: none;
   padding: 14px 28px;
-  background: $Gradient01;
+  background-image: $Gradient08;
```
dkent600, line 6: Can we make the animation be an optional behavior?

First, are you sure that all three button types are supposed to support the animation?  My bet would be only Primary.  You have only shown the Primary one in the screen shot in this PR.

In any case, I would like to be able to display any and all button types without an animation, simply by changing a property on the button, like a boolean `animate-bk`.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
giladt: Quick question- by animation, do you refer to the behavior of the button or the transition effect?
i.E. - 
For the primary, the behavior is changing the gradient bg. 
For the secondary and tertiary it changes the bg color.
on click all translate 1px up.

Should these behaviors be disabled by the new property or only the smooth transition animation?
I've added a property `do-not-animate`, which currently disable all user interactions.

The primary hover animation introduced here is requested in the design system Figma:
image

That being said- The `spinning loading icon` behavior is still absent.

PR #29, opened 2021-12-26
Introduce `pcard`

giladt:
`pcard` is a reusable component for cards. Can be used for every card element, modals, and/or popups.

Simple:
image

Type "default":
image

Type "Success" | "Warning" | "Alert":
image

Each card contains 3 optional slots:
`header`, `body`, and/or `footer`.
`header` may contain the folowing elements:
`<span class="text">Header Text</span>` for heading.
`<i class="icon fas fa-times"></i>` for icons.

image

### DIFF
--- a/src/resources/elements/primeDesignSystem/pcard/pcard.ts
+++ b/src/resources/elements/primeDesignSystem/pcard/pcard.ts
@@ -1,0 +1,20 @@
+import { bindable } from "aurelia-typed-observable-plugin";
+import { customElement } from "aurelia-framework";
+import "./pcard.scss";
+
+/**
+ * Usage:
+ *    <pcard>Default</pcard>
+ *    <pcard type="success">Success</pcard>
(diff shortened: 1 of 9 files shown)


src/resources/elements/primeDesignSystem/pcard/pcard.ts
```diff
@@ -0,0 +1,20 @@
...
+ *    <pcard type="success">Success</pcard>
+ *    <pcard type="alert">Alert</pcard>
+ *    <pcard type="warning">Warning</pcard>
+ *    <pcard ... click.delegate="message('Hi!')">Clickable</pcard>
+ *    <pcard ... width="100%">Fix Size</pcard>
+*/
+export type CardType = "default" | "success" | "alert" | "warning" | "";
+
+@customElement("pcard")
+export class PCurd {
```
dkent600, line 17: PCard

dkent600 (changes requested):

PR #31, opened 2022-01-01
implement css-reset

giladt:
Implemented CSS-Reset package: `the-new-css-reset`.
https://www.npmjs.com/package/the-new-css-reset (v1.4.4)

> A reset stylesheet (or CSS reset) is a collection of CSS rules used to clear the browser's default formatting of HTML elements, removing potential inconsistencies between different browsers.

This eliminates the need to use only `div`s in order to get expected behavior from every HTML element and allows the use of semantic HTML which can better comply with screen readers, and other accessibility tools as well as better SEO results. It would eventually reduce coding time and bugs by presenting cross-browser consistency.


dkent600 (changes requested):
It bugs me that the prime system elements will rely on this .  What if we add a reademe.md under 
 primeDesignSystem that makes clear that it was developed assuming the reset package is present?

dkent600 (changes requested):
need that readme

dkent600:
need to fix the merge conflicts

dkent600 (approved):

PR #32, opened 2022-01-01
Introduce `pcard`

giladt:
# `pcard` is a reusable component for card styling.

## Demo Page:
https://prime-deals-dapp-git-design-system-pcard-curvelabs.vercel.app/storybook/pcard

Simple:

Type Gradient: `<pcard type="gradient">

With styled content:


src/home/home.html
```diff
@@ -6,27 +6,27 @@
...
       </div>
-      <div class="buttons">
-        <div class="buttonContainer">
-          <div class="contribute" click.delegate="navigate('deals')">
-            <div class="title">
-              <div class="header">All Deals</div>
-              <div class="arrow">&#8599;</div>
-            </div>
+      <div class="cards">
+        <pcard type="default" width="360px" click.delegate="navigate('deals')">
```
dkent600, line 10: why type="default"?  just make that the default in the pcard code.  It's not really a default if you have to specify "default"

src/resources/elements/primeDesignSystem/pcard/pcard.scss
```diff
@@ -0,0 +1,114 @@
+@import "styles/styles.scss";
```
dkent600, line 1: the prime system must not have dependencies from outside of itself (such as styles.scss).  Need to remove this.

src/resources/elements/primeDesignSystem/pcard/pcard.scss
```diff
@@ -0,0 +1,114 @@
...
+      font-weight: 900; 
+      margin-right: 6px;
+    }
+  }
+
+  &.default {
+    border-image: linear-gradient(to bottom, $Primary01, $Secondary02) 1 100%;
+  }
+  &.success {
+    border-image: linear-gradient(to bottom, $Green, $Green) 1 100%;
```
dkent600, line 94: We should only be using colors from the design system

src/resources/elements/primeDesignSystem/pcard/pcard.scss
```diff
@@ -0,0 +1,114 @@
...
+    color: $White;
+    margin-bottom: 16px;
+    gap: 6px;
+    >span, >div, >i {
+      width: -webkit-fill-available;
+    }
+
+    .text {
+      font-weight: bold;
+      font-family: "Aeonik";
```
dkent600, line 30: Hmmm, I think the design system needs to specify the font families itself as variables in its own _styles.scss

dkent600 (changes requested):

src/home/home.scss
```diff
@@ -43,61 +42,39 @@
...
-          width: 300px;
-
-          padding: 18px;
-
-          .title {
+          header {
             display: flex;
+            font-family: "Aeonik";
+            font-weight: 700;
+            font-size: 28px;
```
hiaux0, line 56: What's the reason for not having these in the _variables.scss?

src/resources/elements/primeDesignSystem/pcard/pcard.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Usage:
+ *    <pcard>Default</pcard>
+ *    <pcard type="success">Success</pcard>
+ *    <pcard type="alert">Alert</pcard>
+ *    <pcard type="warning">Warning</pcard>
+ *    <pcard ... click.delegate="message('Hi!')">Clickable</pcard>
+ *    <pcard ... width="100%">Fix Size</pcard>
+*/
+export type CardType = "gradient" | "success" | "alert" | "warning" | "";
```
dkent600, line 14: We aren't doing success, alert, warning.  Should remove these

src/styles/styles.scss
```diff
@@ -42,6 +42,7 @@ $BodyPaddingLeftRightMobile: calc((100% - #{$ContentBodyWidthMobile}) / 2);
 /**
  * this is to sync with when the navbar controls start to break
  */
+$PageBreakWidthWide: 1080px;
```
dkent600, line 45: Why here?  This looks like a home-page-specific thing.

dkent600 (changes requested):

src/styles/styles.scss
giladt: removed

hiaux0: I understood the intention of Doug's comment to only move it to another place.
We probably need this variable in other places too

src/resources/elements/primeDesignSystem/pcard/pcard.scss
```diff
@@ -0,0 +1,20 @@
+@import "../styles";
+
+.pcard {
+  padding: $spacing-normal; // only if there is no padding on the parent
+  border-radius: $border-radius-small;
+  cursor: pointer;
```
hiaux0, line 6: I'm wondering about the default state of `cursor: pointer`.
My initial reaction when checking out the demo, was "Why does this have a pointer"?

What are your thoughts on this @giladt ?

src/resources/elements/dealSummary/dealSummary.html
```diff
@@ -45,4 +45,4 @@ <h2>${deal.title}</h2>
     </div>
 
   </div>
-</template>
+</template>
```
hiaux0, line 48: @giladt @dkent600 @paweljanicki @IonelLupu 
We should agree on adding new line or not

Add: 👍 
No New line 🚀

src/resources/elements/primeDesignSystem/pcard/pcard.scss
giladt: good catch. It's a leftover from the prev version. But I agree it shouldn't be defined by the component, but rather by the actual use of it.

src/resources/elements/dealSummary/dealSummary.html
giladt: I have no opinion yet. When/why does it matter?

hiaux0: It matters, because otherwise every PR potentially adds/removes, thus creates a diff aka. noise for the review.

src/home/home.scss
```diff
@@ -43,61 +42,34 @@
...
             line-height: 20px;
-            font-weight: 400;
-            font-family: "Inter";
+            color: $Neutral03;
           }
         }
+        flex-direction: row;
+        flex-wrap: wrap;
+        justify-content: start;
+        cursor: pointer;
```
hiaux0, line 71: Pointer intended for `.card`?

Nit: I'm suggesting to put css definitions directly under selector (here, it is at the bottom, after all child styles, thus hard to see where it belongs to)

src/resources/elements/dealSummary/dealSummary.html
IonelLupu: We can also enforce that with eslint

dkent600: Yes, must be a rule in eslint

src/home/home.scss
dkent600: agreed about putting at the top

dkent600 (approved):

src/resources/elements/dealSummary/dealSummary.html
dkent600: actually already is a rule in our eslint config

PR #33, opened 2022-01-01
Added `pbutton` hover animation

giladt:
Result:
pbutton animation 

button-animation

button-is-loading

button-is-loading-toggle


giladt:
> @dkent600:
> Can we make the animation be an optional behavior?
> 
> First, are you sure that all three button types are supposed to support the animation? My bet would be only Primary. You have only shown the Primary one in the screen shot in this PR.
> 
> In any case, I would like to be able to display any and all button types without an animation, simply by changing a property on the button, like a boolean animate-bk.

Quick question- by animation, do you refer to the behavior of the button or the transition effect?
i.E. -
For the primary, the behavior is changing the gradient bg.
For the secondary and tertiary it changes the bg color.
on click all translate 1px up.

Should these behaviors be disabled by the new property or only the smooth transition animation?
I've added a property do-not-animate, which currently disable all user interactions.

The primary hover animation introduced here is requested in the design system Figma:
image

That being said- The spinning loading icon behavior is still absent now added.Quick question- by animation, do you refer to the behavior of the button or the transition effect?
i.E. -
For the primary, the behavior is changing the gradient bg.
For the secondary and tertiary it changes the bg color.
on click all translate 1px up.

Should these behaviors be disabled by the new property or only the smooth transition animation?
I've added a property do-not-animate, which currently disable all user interactions.

The primary hover animation introduced here is requested in the design system Figma:
image

That being said- The spinning loading icon behavior is still absent now added.

src/resources/elements/primeDesignSystem/pbutton/pbutton.html
```diff
@@ -1,5 +1,6 @@
 <template>
-  <button class="${type} ${fullWidth?'full-width':''}" disabled.bind="disabled" role="button">
+  <button class="${type} ${fullWidth?'full-width':''} ${doNotAnimate? 'no-animation': ''}" disabled.bind="disabled" role="button">
     <slot></slot>
+    <i if.bind="isLoading" class="spinner ${isLoading? 'is-loading': ''} fas fa-sync-alt"></i>
```
dkent600, line 4: uh, this is not part of animation, but ok

src/resources/elements/primeDesignSystem/pbutton/pbutton.ts
```diff
@@ -19,5 +19,7 @@ export type ButtonType = "primary" | "secondary" | "tertiary";
 export class PButton {
   @bindable.string type: ButtonType;
   @bindable.booleanAttr disabled = false;
+  @bindable.booleanAttr doNotAnimate = false;
```
dkent600, line 22: can you please confirm whether secondary and tertiary are supposed to have animation?  They don't have it in your screenshots, but you seem to have been coded in the CSS to assume they do.

Be sure to to document the use cases in the viewmodel

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/pbutton/pbutton.ts
giladt: I've changed the `no-animation` to effct only the primary color transition. It currently means that there is no visable "hover" state for primary buttons if `no-animation` is selected.

`no-animation` have no effect over "secondary" and "tertiary".

dkent600 (approved):

PR #39, opened 2022-01-03
Draft update for readme.md file

giladt:
updating the `readme.md` file.


README.md
```diff
@@ -1,113 +1,101 @@
...
 
-This is the [PrimeDAO](primedao.eth.link) web interface enabling the creation of and interaction with IDOs.
+This prime-deals-dapp repository contains in its master branch the web user interface for the Prime Deals website that is deployed to https://deals.prime.xyz.
 
-## Technical Description
+## Development
 
-This project is bootstrapped by [aurelia-cli](https://github.com/aurelia/cli).
+### Prerequisites
+Make sure you have [node.js version >= 14.11.0 < 15](https://nodejs.org/en/)
```
dkent600, line 9: This is not what the package.json file says.  Can you update that as part of this PR?

dkent600: Should be < 15, or <= 15?

README.md
```diff
@@ -1,113 +1,101 @@
...
-For more information about bundling, go to https://aurelia.io/docs/cli/webpack
-
-## Install
+### Install
 Install dependencies with the following command:
 ```
 npm ci
 ```
+### Update Contract ABIs
+Prime Deals relies on solidity contract addresses and ABIs that it obtains from the [TODO: ADD LINK TO CONTRACTS REPO](https://github.com/PrimeDAO/). You must clone the contracts repository in a folder sibling to this one.
```
dkent600, line 17: this will be our existing contracts-v2

README.md
```diff
@@ -1,113 +1,101 @@
...
+PINATA_API_KEY_TEST=...
+PINATA_SECRET_API_KEY_TEST=...
 
-### Build and serve, running against kovan
-`npm run serve-dev`
+IPFS_GATEWAY="https://${hash}.${protocol}.dweb.link/"
 
-### Build and serve, running against mainnet
-`npm run serve-dev-mainnet`
+NETWORK=rinkeby // or mainnet
```
dkent600, line 40: I wouldn't include NETWORK and PORT here because they aren't concerned about vercel

README.md
```diff
@@ -1,113 +1,101 @@
...
-`npm run build-dev-mainnet`
+### Build
+The package.json file contains lots of commands for building or serving up the application.
 
-## Debug
+#### Environment
+Before building, make sure to have the following in your OS environment variables or in an ".env" file:
+```jsonc
+RIVET_ID=...
+INFURA_ID=...
```
dkent600, line 32: INFURA_ID not required

README.md
dkent600: Should include "COINGECKO_API_KEY="

README.md
```diff
@@ -1,113 +1,101 @@
...
+ETHPLORER_KEY=...
```
dkent600, line 33: ETHPLORER not required

README.md
dkent600: Missing
PINATA_API_KEY=
PINATA_SECRET_API_KEY=

dkent600 (changes requested):

README.md
giladt: > Should be < 15, or <= 15?

Following @paweljanicki proposal, it should be `<15`, so no major upgrades for node will be done.

README.md
giladt: > I wouldn't include NETWORK and PORT here because they aren't concerned about vercel

I disagree regarding the NETWORK variable. When someone would try to deploy the app, NETWORK is required, either in the platform environment variables definitions or the run command. But it needs to be clear that the variable needs to be passed.

dkent600: Thing is, the variable does not "need" to be passed.  Note in main.ts it defaults to rinkeby or mainnet, depending on whether is a development or production build.  So really, the only concern here is Vercel, and that is only *our* concern. We are not concerned here about how other people might deploy the app to their own idiosyncratic servers.  I still think should leave NETWORK out.

dkent600 (changes requested):
Still need to remove yarn.lock

README.md
giladt: Oh, fair point. Didn't see that defaulting in main.js.

dkent600 (approved):

PR #55, opened 2022-01-05
Chore: Add cypress

hiaux0:
## What was done
Copy over aurelia-cli generated Cypress Stuff
[Some notes on Notion](https://www.notion.so/primedao/Front-End-Coding-Guidelines-a18c230471294a9fb4ad3e570468ce00#a044ea0d93964bd2876e8caba6f860d3)

Note, this is only setup with no real functionality

## Run
1. `npm run e2e`
2. A Cypress window opens

3. click on `app.e2e.ts`
4. A test runner opens
  - left side: test info like, commands or requests
  - right side: UI, that the commands act upon
  - (I used Firefox in the screenshot)


hiaux0:
- [ ] Fix sourcemaps
  It said to add `devtools: 'inline-source-maps', but was still not as expected

PR #56, opened 2022-01-06
fix(*): conosle errors (`this.ipfs` and `deal.daos` undefined)

hiaux0:
## What was done

1. `this.ipfs` undefined --> fix by changing initialize order in `main.ts`
2. `deal.daos` undefined --> fix by if-guard

## Why?
1. For e2e testing setup, this was failing tests
2. Spammed the console with error logs


src/services/DealService.ts
```diff
@@ -73,6 +73,7 @@ export class DealService {
     hashes.forEach( async (hash:string) => {
       this.dealsObject[hash] = await (this.ipfsService.getObjectFromHash(hash) as Promise<IDealConfig>)
         .then(async (deal: any) => {
+          if (deal.daos === undefined) return {};
```
hiaux0, line 76: This is rather a quick fix.
Potential issue: Caller place expects a `Deal` object, that has nested fields.

"Clean fix" suggestion: 
1. return undefined here
2. In caller place if-guard for undefined (`if (result === undefined) return`)

dkent600 (approved):

PR #57, opened 2022-01-07
Remove unwanted horizontal scrollbar

paweljanicki:
Considering how `minmax()` works and the values that were supplied, it was returning value never smaller than 160px. What was causing a horizontal scrollbar, when view port width was smaller that 1440px (160px + 1120px + 160px).
I don't see why we needed that `minmax()` over there as all we wanted was to center the content. `1fr` should be sufficient

### DIFF
--- a/src/app.scss
+++ b/src/app.scss
@@ -4,7 +4,7 @@
 
 .appWrapper {
   display: grid;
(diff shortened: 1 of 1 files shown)


dkent600 (commented):
Doesn't work for all cases, leaves no margin.  That is why I did the minmax.  I am going to go ahead and fix this myself, probably be altering the breakpoints,  Thanks for your suggestion though.

image

PR #58, opened 2022-01-07
Introduce `pselect`

giladt:
# `pselect`- dropdown

Options Data:
```
{
  value: string,
  text: string,
  innerHTML: string
}
```

Optional Parameters: 
`is-searchable` - enables search box in dropdown. Default: false
`placeholder` - placeholder text for initial state. Default: 'Please Select...'

select-dropdown


giladt:
Created a new PR with resolved merge conflicts.

src/resources/elements/primeDesignSystem/pselect/pselect.scss
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  .ss-open-below + .ss-content {
+    border-top-color: $Border01;
+    border-radius: 0 0 10px 10px;
+  }
+  .ss-open-above + .ss-content {
+    border-bottom-color: $Border01;
+    border-radius: 10px 10px 0 0;
```
hiaux0, line 115: General question:
Will we be using more scss variables instead of hard coding values?

giladt: Good point. I think it makes sense to use variables in this case.

dkent600 (approved):

PR #65, opened 2022-01-09
pinput-numeric component

IonelLupu:
Preview: https://prime-deals-dapp-git-pinput-numeric-curvelabs.vercel.app/storybook/pinput-numeric

image


giladt:
Comments regarding styling:
- Inactive: Border color should be `$Border01`
- Disabled: Font color should be `$Secondary05`
- Validating: Border color should be `$Neutral04`
- I'm missing the `error`/`warn`/`validation` labels below the input field
- Placeholder text color should be `$Neutral02` (without opacity)
- Text color should be `$Neutral04`

IonelLupu:
@giladt I've updated the styles. Please let me know if everything is ok.

About the `error/warn/validation labels below `. Since these messages should be present for also textareas, select boxes and any other inputs (eg. when the form is invalid), I was thinking we create a `form-item` component that will take care of displaying these messages. Something like:

```html
<form-item>
    <label>My field</label>
    <pinput-numeric></pinput-numeric>
    <p class="form-error" if.bind="form.myField.errors.required">'My Field' is required</p>
</form-item>

```

or 

```html
      <form-item>
        <form-label>My Field</form-label>
        <form-control errorTip="'My Field' is required">
              <pinput-numeric></pinput-numeric>
        </form-control>
      </form-item>
```

I was inspired by these libraries that I worked with in the past: 
- https://getbootstrap.com/docs/5.0/forms/validation/
- https://ng.ant.design/components/form/en#components-form-demo-auto-tips
- https://v1.tailwindcss.com/components/forms

Not sure why they did this, but I think it was done to separate the concerns between individual inputs and forms. (who knows...)

We can talk about this with the team to see if we:
1. should add the error message functionality inside each and every input (text, textarea, numeric, dropdown, toggle, date etc) 

or 

2. create a form-item component that will take care of labels and errors.

Let me know what you think

giladt:
I like your path of thinking. I believe option 2 is much cleaner, and also lets us reuse the input elements if needed. I believe @dkent600 will agree about that too.

src/resources/elements/primeDesignSystem/_styles.scss
```diff
@@ -1,2 +1,58 @@
...
+  line-height: 23px;
+  border: 1px solid $Border01;
+  background: $BG04;
+  color: $Neutral04;
+  border-radius: $border-radius-normal;
+  outline: none;
+  padding: $spacing-normal;
+  box-sizing: border-box;
+
+  &, input::placeholder {
```
giladt, line 17: font color for filled value is not applied correctly. I guess you should move the decleration into it's underling input element:
```scss
  &>input{
    color: $Neutral04;
    ::placeholder {
      color: $Neutral02;
    }
  }

```

giladt (changes requested):

src/resources/elements/primeDesignSystem/_styles.scss
IonelLupu: Missed that. Thank you! 😄

giladt (approved):

hiaux0:
2cents

> I like your path of thinking. I believe option 2 is much cleaner, and also lets us reuse the input elements if needed. I believe @dkent600 will agree about that too.

I prefer the first option (use as much of existing html tags as possible).
`<label>` has a lot of built-in functionality in the HTML specs, and by implementing our own `<form-label>` we basically have to reimplement those.
In the future, we _may_ be required to add a `<form-label>` component, but until then I'd vote for sticking to vanilla.

Ant, _does_ create their own label component, but Tailwind comes out fine with vanilla.

giladt:
I think the idea was not necessarily creating a custom label, but more creating a parent element for the input which contains a label and a controller. @IonelLupu, correct me if I'm wrong.

IonelLupu:
@hiaux0 @giladt Actually, I am ok with any of those two options. Let's see what @dkent600 has to say about it.

One thing we need to take care of is the space beneath the input for displaying errors. 
1. If we chose to have a space that is always present (in order to make space for the error messages), it will mean that every input will have some sort of margin beneath and it might affect pages that require inputs without that margin. Or maybe we will be lucky enough to not have such requirements.
2. If we choose to have the error show up only when it's needed, without having a space for it by default, it will shift the layout visually up or down making the UX a bit unpleasant.

Or maybe I am missing something and there is a 3rd option I don't know about😨 . Please help me 🤕

src/resources/elements/numericInput/numericInput.ts
```diff
@@ -14,6 +14,7 @@ export class NumericInput {
   @bindable.booleanAttr public decimal = true;
   @bindable public css?: string;
   @bindable({ defaultBindingMode: bindingMode.oneTime }) public id?: string;
+  @bindable(({ defaultBindingMode: bindingMode.twoWay })) inFocus = false;
```
dkent600, line 17: why is inFocus needed?

src/resources/elements/primeDesignSystem/_variables.scss
```diff
@@ -9,7 +9,7 @@ $spacing-small: 5px;
 $spacing-medium-small: 7.5px;
 $spacing-normal: 10px;
 
-$input-height: 40px;
+$input-height: 46px;
```
dkent600, line 12: did you pass this by gilad?  I believe he put this in there to support the pinput-select

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.scss
```diff
@@ -0,0 +1,19 @@
+@import "../styles";
+
+pinput-numeric {
```
dkent600, line 3: are you sure some of this won't also be needed by the other pinput components?

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.scss
```diff
@@ -0,0 +1,19 @@
...
+  @extend .pInput;
```
dkent600, line 4: There should always be a containing element (as I requested elsewhere), and everything in this file should be contained within it.  This ensures good scoping.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/_variables.scss
giladt: I was wondering too, but apparently, the size has changed on Figma.

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.scss
IonelLupu: isn't the component itself a container which also preserves the scope? This is why I used the component tag name as the selector.

src/resources/elements/primeDesignSystem/_variables.scss
IonelLupu: Yes, I took that from the Figma

src/resources/elements/numericInput/numericInput.ts
IonelLupu: I added this just to make the input focus when you use the "inFocus" attribute on the other "pinput-numeric" component. The `pinput-numeric` has a specific border color when focus. And when it's in focus you should also have the focus in the input itself. The "focus" border color alone didn't make sense without having an input which is not in focus.

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.html
```diff
@@ -0,0 +1,28 @@
+<template
+  class="
```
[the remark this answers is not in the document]
IonelLupu: The only reason I added the classes on the template is just to make those classes show up on the component itself and not use an additional tag to do the work since it looked redundant to me. 

Eg: 
```html
<pinput-numeric class="pInput-error">
    ....
</pinput-numeric>
```
VS an additional element like:
```html
<pinput-numeric>
    <div class="pInputContainer pInput-error">
        ....
    </div>
</pinput-numeric>
```

But if that's the way you create components in this project, I will change it. Didn't know that.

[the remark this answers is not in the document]
dkent600: I was not aware that would work.  OK, keep it.

dkent600 (approved):

PR #66, opened 2022-01-10
pinput-text component

IonelLupu:
Preview link: https://prime-deals-dapp-git-pinput-text-curvelabs.vercel.app/storybook/pinput-text
image


hiaux0 (commented):
Code: Looks good and clean
Functionality: I noticed `cursor: default` in between input and the border, imo inside border should be `cursor: text` everywhere

IonelLupu:
> Functionality: I noticed cursor: default in between input and the border, imo inside border should be cursor: text everywhere

@hiaux0  good catch. Will update that

hiaux0 (commented):
Padding issue fixed. Good to go from my side

src/resources/elements/primeDesignSystem/_styles.scss
```diff
@@ -14,39 +14,37 @@
...
     color: $Neutral04;
+    height: 100%;
+    padding: $spacing-normal;
 
     &::placeholder {
       color: $Neutral02;
     }
   }
 
+  .fas {
```
dkent600, line 34: I prefer that we apply and select our own class that applies specifically to the element(s) being styled rather than referencing a more generic and third-party class.

src/resources/elements/primeDesignSystem/pinput-text/pinput-text.html
```diff
@@ -0,0 +1,27 @@
...
+          : ''
+  }
+  ${disabled ? 'pInput-disabled' : ''}
+  ${inFocus ? 'pInput-focused' : ''}
+">
+
+  <input
+    disabled.bind="disabled"
+    value.bind="value"
+    focus.bind="inFocus"
```
dkent600, line 16: do we really need this bindable property in this case?

dkent600 (changes requested):

giladt (approved):

dkent600 (approved):

PR #67, opened 2022-01-10
pinput-textarea component

IonelLupu:


hiaux0:
Why do we call this "input-textarea" instead of just "textarea"?
Might be a good idea to also included links to Figma in design component PRs

IonelLupu:
hi @hiaux0 . There was no particular reason. But you are right, just "textarea" sounds simpler. 

I didn't include any details in this PR because it's still a draft. It's blocked by the PR with the numeric-input. I will add all the details when I will publish this PR. 

Actually, I will add `[WIP]` in the name of the PR

hiaux0:
> Actually, I will add [WIP] in the name of the PR

No need to, the PR being in the `draft` state should already indicate that.
Maybe that comment was in the wrong place. 
I added it, because I saw #65 also not having a link to Figma (which is fine, because we didn't agreed on adding one)

PR #69, opened 2022-01-13
Deals Wizard refactor architecture

paweljanicki:
Ticket: https://github.com/PrimeDAO/prime-deals-dapp/issues/40

New architecture for the deals wizard, using a service for managing wizard.

`DealWizardService` holds the config `IWizardConfig` (stages, what is current stage, result) for each registered wizard, aka `wizardManager`. Also has a method for setting up the routes for the stages and has methods for navigating between stages.

`OpenDealWizard` is a wizard component, aka `wizardManager`. It configures the wizard, currently holding 2 dummy stages.

`OpenDealWizardResult` is a class holding the result from the wizard above. Note: it's passed to the service and access via the service. This class implements `IWizardResult` interface, a basic for any wizard result class. Every wizard will have its own wizard result class, and that class should have wizard specific data structure.

`ProposalStage` is an example of a stage component, a route. The component is not tight to any wizardManager, such as `OpenDealWizard`. I can be reused by multiple wizardManagers, the information about which wizardManager it belongs to is passed by the router `settings`.
It assumes that its data lives under `proposal` property in the wizardResult. In other words a class holding a result of a wizard, such as `OpenDealWizardResult` in case of `OpenDealWizard` should have `proposal` property to correctly use `ProposalStage`. I'm open to discussion on how to improve this

New dummy route, where you can test it out https://prime-deals-dapp-git-feature-registration-wizard-curvelabs.vercel.app/initiate2/open/stage1


src/dealWizard/components/dealWizardStepper/dealWizardStepper.html
```diff
@@ -0,0 +1,17 @@
+<template>
```
dkent600, line 1: need to turn this into a design system component

src/dealWizard/components/dealWizardStepper/dealWizardStepper.html
```diff
@@ -0,0 +1,17 @@
...
+  <div class="dealWizardStepper">
```
dkent600, line 2: being a design system thing, we should remove references to "deal"

src/dealWizard/dealWizard.types.ts
```diff
@@ -0,0 +1,90 @@
+import { IWizardConfig } from "services/DealWizardService";
```
dkent600, line 1: can we make this service, and all its types, agnostic to deals?  so can apply to any wizard?

src/dealWizard/dealWizard.types.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export interface IBaseWizardStage {
+  wizardManager: any;
+  wizard: IWizardConfig;
+  errors: {[key: string]: string};
+
+  activate;
+  attached: () => void;
+  validateInputs(): boolean;
+  proceed: () => void;
```
dkent600, line 87: we don't want to be implementing this stuff in every stage viewmodel.  we should be using the service directly.  the only thing I recall we need in every stage viewmodel is the wizardManager.

src/dealWizard/openDealWizard/openDealWizard.ts
```diff
@@ -0,0 +1,36 @@
...
+  }];
+
+  private wizardResult = new OpenDealWizardResult();
+
+  constructor(private dealWizardService: DealWizardService) {
+    this.dealWizardService.registerWizard(this, this.stages, this.wizardResult);
+  }
+
+  private attached(): void {
+    this.dealWizardService.updateIndexOfActiveBaseOnRoute(this);
```
dkent600, line 30: doesn't this have to be called when attaching to each stage?

src/services/DealWizardService.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  public goToStage(wizardManager: any, index: number): void {
+    const wizard = this.getWizard(wizardManager);
+    this.router.navigate(wizard.stages[index].route);
+    wizard.indexOfActive = index;
+  }
+
+  public updateIndexOfActiveBaseOnRoute(wizardManager: any): void {
```
dkent600, line 90: can you document when this needs to be invoked?

src/services/DealWizardService.ts
```diff
@@ -0,0 +1,96 @@
...
+        wizardManager: wizardManager,
+      },
+    }));
+
+    config.map(routes);
+
+    this.router = router;
+  }
+
+  public getWizard(wizardManager: any): IWizardConfig {
```
dkent600, line 53: can we be consistent with calling this either "wizard" or "wizardConfig"?

src/dealWizard/stages/primaryDao/primaryDao.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  validateInputs(): boolean {
+    this.errors = {};
+
+    if (!this.wizard.wizardResult.daos[0].name) {
+      this.errors.name ="Please enter the name of the Primary DAO";
+    }
+
+    this.wizard.stages.find(stage => stage.name === "Primary DAO").valid = !Object.keys(this.errors).length;
```
dkent600, line 30: why not call ` getWizardStage`?

the `currentStage` can be a method on the service.  I think we can maintain this in the router logic.

src/dealWizard/openDealWizard/openDealWizard.ts
dkent600: we should maintain this state in the service.  I think we can do this in the routing logic.

src/dealWizard/stages/primaryDao/primaryDao.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    return !Object.keys(this.errors).length;
+  }
+
+  proceed(): void {
```
dkent600, line 35: as mentioned, would be better to call directly into the service rather than implement these in every stage.

dkent600 (changes requested):

IonelLupu:
Typescript is very good at catching errors if used in strict mode (`strict: true` in tsconfig.json) . I know this is not enforced yet in the project but I suggest we activate it in the future. Meanwhile, we can prepare our new code to follow the "strict" rules of typescript.

Having this said, I recommend you typehint every variable, property and argument to something more specific and remove the `any` keyword. 😄

src/dealWizard/openDealWizard/openDealWizardResult.ts
```diff
@@ -0,0 +1,54 @@
...
+        }],
+        social_medias: [{name: undefined, url: undefined}],
+        logo_url: null,
+      },
+      {
+        name: "",
+        tokens: [{name: undefined, amount: undefined}],
+        social_medias: [{name: undefined, url: undefined}],
+      },
+    ] as IDAO[];
```
IonelLupu, line 41: is `as IDAO[]` necessary? I see `this.daos` is already of type `IDAO[]`

src/dealWizard/openDealWizard/openDealWizardResult.ts
```diff
@@ -0,0 +1,54 @@
...
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    this.terms = {
+      clauses: [{text: "", tag: ""}] as IClause[],
```
IonelLupu, line 44: same here as well. is `as IClause[]` required. I think it can be removed. 

Same for the `as`es below 😄

src/dealWizard/stages/primaryDao/primaryDao.ts
```diff
@@ -0,0 +1,40 @@
...
+import { RouteConfig } from "aurelia-router";
+import { WizardService, IWizard } from "services/WizardService";
+import { IBaseWizardStage } from "../../dealWizard.types";
+import "../wizardStage.scss";
+
+@autoinject
+export class PrimaryDAO implements IBaseWizardStage {
+  public wizardManager: any;
+  public wizard: IWizard;
+  public errors: {[key: string]: string} = {};
```
IonelLupu, line 11: You can use `Record<string, string>` instead of a custom object type. `Record` is widely used in typescript programs:
https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

IonelLupu (changes requested):

src/dealWizard/stages/proposalStage/proposalStage.ts
```diff
@@ -31,22 +31,14 @@ export class ProposalStage implements IBaseWizardStage {
       this.errors.overview = "Please enter a descriptive overview for your proposal";
     }
 
-    this.wizard.stages.find(stage => stage.name === "Proposal").valid = !Object.keys(this.errors).length;
+    this.wizardService.getCurrentStage(this.wizardManager).valid = !Object.keys(this.errors).length;
 
     return !Object.keys(this.errors).length;
   }
 
   proceed(): void {
```
dkent600, line 39: We can move this to the service.  In the HTML, just call `proceed (validateInputs.bind())` and let the service do the work.

src/dealWizard/dealWizard.types.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface ITerms {
+  clauses: Array<IClause>,
+  period: number,
+  representatives: string,
+  coreTeamChatURL: string,
+  previousDiscussionURL: string,
+}
+
+export interface IBaseWizardStage {
```
dkent600, line 61: My take:  I am not sure this interface is needed for anything.  Stages can implement themselves however they want, as long as they get the job done.

src/dealWizard/stages/primaryDao/primaryDao.html
```diff
@@ -0,0 +1,37 @@
...
+            <label for="proposalName">DAO Name</label>
+            <span>
+              ${wizard.wizardResult.daos[0].name.length}/50
+            </span>
+          </div>
+          <input id="proposalName" value.bind="wizard.wizardResult.daos[0].name" maxlength="50" />
+          <span>${errors.name}</span>
+        </div>
+        <div class="buttonContainer">
+          <pbutton type="secondary" click.delegate="wizardService.previous(wizardManager)">
```
dkent600, line 24: I am tempted to suggest we turn these buttons into reusable components where we just pass in the wizardService.

Maybe could have a component that contains all of them and displays the exact required buttons for first, middle and last stages.

dkent600 (changes requested):

PR #80, opened 2022-01-20
Pform input display block + textarea resize

IonelLupu:


paweljanicki (approved):

src/resources/elements/primeDesignSystem/ptextarea/ptextarea.ts
```diff
@@ -7,11 +7,12 @@ import {ValidationState} from "../types";
...
-  @bindable.booleanAttr public disabled;
-  @bindable.number public rows = 4;
-  @bindable({defaultBindingMode: bindingMode.twoWay}) public value: string | BigNumber;
-  @bindable.string public placeholder = "";
+  @bindable autocomplete = "off";
+  @bindable.booleanAttr disabled;
+  @bindable.number rows = 4;
+  @bindable({defaultBindingMode: bindingMode.twoWay}) value: string | BigNumber;
+  @bindable.string placeholder = "";
+  @bindable.booleanAttr resizable = true;
```
dkent600, line 17: Test this carefully.  In my experience `@bindable.booleanAttr` only works the way we expect if its default value is `false`

src/resources/elements/primeDesignSystem/ptextarea/ptextarea.ts
```diff
@@ -7,11 +7,12 @@ import {ValidationState} from "../types";
...
-  @bindable public validationState?: ValidationState;
+  @bindable validationState?: ValidationState;
 
-  @bindable public autocomplete = "off";
-  @bindable.booleanAttr public disabled;
-  @bindable.number public rows = 4;
-  @bindable({defaultBindingMode: bindingMode.twoWay}) public value: string | BigNumber;
-  @bindable.string public placeholder = "";
+  @bindable autocomplete = "off";
+  @bindable.booleanAttr disabled;
```
dkent600, line 13: best to explicitly set default to `false`

src/resources/elements/primeDesignSystem/_styles.scss
```diff
@@ -71,7 +71,7 @@
 
   .feedback-icon {
     right: $spacing-normal;
-    top: 35%;
+    top: calc(50% - 8px); // centers the icon
```
dkent600, line 74: what is the purpose of  `.pInput-parent`?  Why is it in styles.sscss?

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/_styles.scss
IonelLupu: The purpose of this class is to add it on the input components we create (`pinput-numeric`, `pinput-text` etc) which allows us to center the validation icons and color them. 

You may say that we can style the validation icons using the `.pInput` class but this is not possible in the case of the `pinput-numeric` because the actual input is inside another component caller `numeric-input`.

I'm open to suggestions  to move this inside the `.pInput` class 😄  @dkent600  @hiaux0 @giladt  @paweljanicki

src/resources/elements/primeDesignSystem/ptextarea/ptextarea.ts
IonelLupu: me and @paweljanicki tested this on the [storybook page](https://prime-deals-dapp-git-pform-input-display-block-curvelabs.vercel.app/storybook/ptextarea) and it seems it works as expected. Do you have a specific edge-case we should check for?

src/resources/elements/primeDesignSystem/_styles.scss
hiaux0: Not too familiar with structure of scss right now, but `.pInput-parent` in `_styles.scss` does seem too specific.
What's stopping us from creating sth like "common-input.scss" (read: shared or common place to put input stuff)

IonelLupu: the `.pInput-parent` class is used on the `template` tag like this: `<template class="pInput-parent ${ validationState ? 'pInput-parent-' + validationState : '' }">` (check `pinput-numeric`, `pinput-text` etc). This means that many input components will use it and that's why I added this class in `_styles.scss`. 

Nothing stops us creating `common-input.scss` tho'. Not sure how necessary is to create a component just to scope some SCSS. Your call @hiaux0 @dkent600 . Should I do it?

dkent600: I like putting it in pinput.scss.  Happy to brainstorm on how to do it.

src/resources/elements/primeDesignSystem/ptextarea/ptextarea.ts
dkent600: ok I don't recall, lets leave it

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/_styles.scss
IonelLupu: I moved the input related styles into pinput.scss. What do you think?

dkent600 (approved):

Issue #95, opened 2022-01-25
Implement DealsService and DealEntities


hiaux0:
#81 

> Do this first, because we want to push assumptions about Ceramic to the latest time possible

PR #89, opened 2022-01-24
Updated home page based on the mockup design

IonelLupu:
## What was done
The home page was updated based on the [mockup design](https://www.figma.com/file/hTLvqCEzki3WoG5uJbtynw/Prime-Deals?node-id=0%3A1):
image

#### After

PR #76, opened 2022-01-19
Add 2 wizards with 1st stage and boilerplate for stage 2 and 3

paweljanicki:
LATEST UPDATE: (Already review in the PR https://github.com/PrimeDAO/prime-deals-dapp/pull/86)
- Add "make an offer wizard"
- Add boilerplate for stages "Proposal Lead" and "Primary DAO" and "PartnerDAO"
- Add some components shared across stages
- Passing data to stage components via route settings

UPDATE:
- First stage of the wizard
- stageWrapper component to reuse layout and basic styling between stages (no logic)
- Refactor and improvements to pstepper component

https://prime-deals-dapp-git-feature-wizard-stage-1-curvelabs.vercel.app/initiate/token-swap/open-proposal/stage1

ALREADY REVIEWED:
Two steps for selecting a deal type

Step 1: https://www.figma.com/file/hTLvqCEzki3WoG5uJbtynw/Prime-Deals?node-id=3566%3A70484
Step 2: https://www.figma.com/file/hTLvqCEzki3WoG5uJbtynw/Prime-Deals?node-id=3566%3A70438

I have:
- refactored code from the step one and made it a reusable component
- added e2e tests for selecting deal type
- connected the new "openDealWizard" (WIP) to the flow

PR #96, opened 2022-01-25
 Add DealsService mock methods and DealEntities

hiaux0:
#95
## What was done
This PR is part of #81, and it deals with the Frontend API (CeramicService will come later)

Add
1. DealsService methods
    - 2 methods to get wizard data (cf. #76)
    - 2 methods to get open/partner deals for the home page (cf. #89)
2. DealEntities
    - Stitched together from #76 and #89, will need to align on this
    - cc @paweljanicki and @IonelLupu

### DIFF
--- a/src/services/DealService.ts
+++ b/src/services/DealService.ts
@@ -5,6 +5,8 @@ import { autoinject } from "aurelia-framework";
 import { IDealConfig } from "../registry-wizard/dealConfig";
 import { OPEN_DEALS_MOCK, PARTNERED_DEALS_MOCK } from "entities/MockDealEntity";
 import { IDummyDeal } from "entities/Deal";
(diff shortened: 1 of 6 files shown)


hiaux0:
@dkent600 

You were right in pointing out #77 was more about DealService and DealEntities.
I closed 77 in favor for this PR, because makes sense to do Service and Entities first.

As outlined in #81, I will tackle the DataStorageMock afterwards

src/services/DealService.ts
```diff
@@ -64,11 +68,31 @@ export class DealService {
     /**
      * deals will take care of themselves on account changes
      */
-    this.getDeals();
-    this.getDAOsInformation();
+    this.getDealsIPFS();
+    // this.getDAOsInformation();
```
hiaux0, line 72: @dkent600 @giladt 

Right now, we always load deepDAO information, imo, we should disable?
Reason: I want to prevent unnecessary calls.

This is a side "fix" and could be done separately

hiaux0:
#97 replaces this one

PR #97, opened 2022-01-25
 Implement mock data storage service

hiaux0:
#81 

## What was done
Most of the work was by @dkent600 
1. `main.ts` to register dependencies and initialize singleton services (`CeramicServiceMock` and `DealService`
2. `DealService.ts` uses `DataSourceDeals` to fetch data from the Data Storage
   1. From the payload `DealEntities` are created and initialized
   - `CeramicServiceMock` is a concrete implementation of `DataSourceDeals` 
   - 
3. `DealEntities` hydrates and makes data avaiable (through getters) to the remaining application

## Todo
- [x] add flow, that simulates ceramic streams


src/main.ts
```diff
@@ -26,6 +29,7 @@ export function configure(aurelia: Aurelia): void {
       configuration.settings.keyboard = false;
     });
   aurelia.use.singleton(HTMLSanitizer, DOMPurify);
+  aurelia.use.singleton(CeramicServiceMock, CeramicServiceMock);
```
hiaux0, line 32:
```ts
export declare class DataSourceDeals extends IDataSourceDeals {}

export declare class DataSourceDeals implements IDataSourceDeals {
  initialize(rootId?: string): void;
  get<T>(id?: string): T;
  create(registration: IDealConfig): Promise<string>;
  save(id: string, registration: IDealConfig): void;
  update(id: string, registration: IDealConfig): void;
}
```

src/main.ts
```diff
@@ -26,6 +29,7 @@ export function configure(aurelia: Aurelia): void {
       configuration.settings.keyboard = false;
     });
   aurelia.use.singleton(HTMLSanitizer, DOMPurify);
+  aurelia.use.singleton(DataSourceDeals, CeramicServiceMock);
```
hiaux0, line 32: @dkent600 
#### Use implementation as "key"
In the pair session, we had
```suggestion
  aurelia.use.singleton(IDataSource, CeramicServiceMock);
```

But this gave an DI error: 
`Error: key/value cannot be null or undefined. Are you trying to inject/register something that doesn't exist with DI?`

So I had to go the way to create an Implementation `IDataSource` in `DataSourceDeals.ts`

#### Other approaches
I tried other approaches with no luck

```ts
export declare class IDataSourceDeals;

export declare class DataSourceDeals extends IDataSourceDeals {}

export declare class DataSourceDeals implements IDataSourceDeals {
  initialize(rootId?: string): void;
  get<T>(id?: string): T;
  create(registration: IDealConfig): Promise<string>;
  save(id: string, registration: IDealConfig): void;
  update(id: string, registration: IDealConfig): void;
}
```

src/entities/Deal.ts
```diff
@@ -21,15 +25,37 @@ export class Deal {
...
+  // public get votes(): Array<IVoteInfo> {
+  //   return this.rootData.votes;
+  // }
+
+  // public get discussions(): Array<Array<IClause, Hash>> {
+  //   return this.rootData.discussions;
+  // }
+
+  public get isOpen(): boolean {
+    return this.rootData?.registration.daos.length === 1;
```
hiaux0, line 41: This is just an assumption I'm making about how to determine "open deals".
Should be replaced when we have a stable schema

dkent600 (approved):
Maybe this has an interesting solution?  @hiaux0 

https://stackoverflow.com/questions/30872744/aurelia-di-with-typescript-interfaces

PR #99, opened 2022-01-26
Trigger wizard stage validation in the wizardService

paweljanicki:
## What was done
- Get rid of validation methods that were passed down the tree of components
- Trigger validation directly in the `wizardService` in `proceed` method.

## Benefits
- One stage (one stage component) could have 2 or more different validation methods based on context (so we could show different errors based on context)
- Validation could be triggered outside of a stage component (for example via stepper)
- We can re-run the validation at “submit” so for example if user correctly fills stage one and click proceed, it is marked as “valid”. However if they come back and edit the stage and later navigate to another stage by clicking the stepper validation is not triggered and result is not updated. If on clicking “submit” on the last stage with don’t re-validate that edited stage, user will think that they submitted their latest edits, while they didn’t. 

#### Before
- Stage validation could not be triggered from outside of the component
#### After
- Stage validation is triggered in the `wizardService` and therefore could be triggered by any component or service that injects `wizardService`


src/dealWizard/stages/partnerDaoStage/partnerDaoStage.html
```diff
@@ -4,7 +4,7 @@
   <stage-wrapper
     title="Partner DAO Details"
     wizard-manager.bind="wizardManager"
-    validate.call="validateInputs()">
+    show-submit.bind="true">
```
hiaux0, line 7: This seems a bit icky.
Validation logic for submit is handled somewhere else now?

_(Maybe all good, I'm just going from the diff here)_

src/services/WizardService.ts
```diff
@@ -8,12 +9,18 @@ export interface IWizardState {
   registrationData: IRegistrationData;
 }
 
-export interface IWizardStage {
+export interface IWizardStage<Errors = any> {
```
hiaux0, line 12: Could be a good practice to always use `unknown` over `any`.
This is what somehow stuck to my mind.

Quick search:
https://stackoverflow.com/questions/51439843/unknown-vs-any

IonelLupu, line 12: Is it possible to typehint all of these additional `any`s? I can help with that. Maybe `Errors` can be an `object`?

src/services/WizardService.ts
```diff
@@ -8,12 +9,18 @@ export interface IWizardState {
...
   name: string;
   valid: boolean;
   route: any;
   moduleId: any
   settings?: {[key: string]: any};
+  errors?: Errors;
+  validationMethod?: any;
```
IonelLupu, line 19: same here with `any`?

IonelLupu (changes requested):

hiaux0 (approved):
## Code
Well done. You have my thumb 👍 

## Functionality
- Validate on blur
    - A quick test, showed, that we don't re-validate on blur. Would that be possible in the future to add that?

image

src/dealWizard/openProposalWizard/openProposalWizardManager.ts
```diff
@@ -13,11 +14,13 @@ export class OpenProposalWizardManager {
     valid: false,
     route: "stage1",
     moduleId: PLATFORM.moduleName("../stages/proposalStage/proposalStage"),
+    validationMethod: this.wizardValidationService.validateProposalStage,
```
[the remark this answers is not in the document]
paweljanicki: The reason why I'm passing it here is because the `validationMethod` is triggered by the `wizardService` not by a stage component.

src/dealWizard/stages/partnerDaoStage/partnerDaoStage.html
paweljanicki: So the `show-submit.bind="true"` line is unrelated to the validation. It just happened that I made this change on this branch. It basically shows submit button instead of proceed button

src/services/WizardService.ts
paweljanicki: Thanks for the hint

paweljanicki:
> ## Code
> Well done. You have my thumb 👍
> 
> ## Functionality
> * Validate on blur
>   
>   * A quick test, showed, that we don't re-validate on blur. Would that be possible in the future to add that?
> 
> image

Yes it is, stage can call validation method, so we could call it on blur

dkent600 (changes requested):
Quick impressions:

1) I like the idea of a service such as this
2) I am more comfortable with defining the validation execution inside the stages themselves.

Can we combine these?

paweljanicki:
> Quick impressions:
> 
> 1. I like the idea of a service such as this
> 2. I am more comfortable with defining the validation execution inside the stages themselves.  This keeps all the stage-specific knowledge in one place
> 
> Can we combine these?

Yes, we can do it. I'll change it

src/dealWizard/makeOfferWizard/makeOfferProposalLeadStage/makeOfferProposalLeadStage.ts
```diff
@@ -20,19 +20,16 @@ export class MakeOfferProposalLeadStage implements IBaseWizardStage {
 
   attached(): void {
     this.wizardState = this.wizardService.getWizardState(this.wizardManager);
+    this.wizardService.registerStageValidateFunction(this.wizardManager, this.validate.bind(this));
```
dkent600, line 23: preference is to use arrow functions rather than bind

dkent600 (approved):

PR #100, opened 2022-01-27
Add index page to storybook

hiaux0:
## What was done
- Add index/welcome demo page
- refactor existing to accommodate for hat
image

## How to add new demo
1. In `src\resources\elements\primeDesignSystem\demos\`
2. Add your demo component 
    1. Has to have a View model (because of how `<compose>` works here)
    2. Has to be inside a folder `myDemoComponent/myDemoComponent`
3. In `demosWelcome.html` add your component
    1. `<require>` it
    2. Add a button to link to it

## Todo
- [ ] Find a way to not having to add view models

#### Before
- have to remember url to storybook elements

#### After
- index page

## Why was this done
For the CeramicService work, I wanted to have a place in the app, where I can test things.
In other places, I would have add a `playground` route, but decided to just create a temporary "demo component" in the `/storybook` route.

I believe this refactor will be beneficial for us in the future, because it gives an overview of all components.


src/resources/elements/primeDesignSystem/demos/demos.ts
```diff
@@ -1,86 +1,19 @@
...
-import {Router, RouterConfiguration} from "aurelia-router";
+import {Router} from "aurelia-router";
 
 export class Demos {
 
   router: Router;
 
-  private configureRouter(config: RouterConfiguration, router: Router) {
-    const routes = [
+  private configureRouter(config, router): void {
```
paweljanicki, line 8: Do we have to get rid of typing the arguments?

paweljanicki (approved):

paweljanicki:
Great work! I like the idea, easier to use the UI and also the code is cleaner

dkent600:
Design system has no dependency on ceramicservice, so I don't understand the need.

hiaux0:
> Design system has no dependency on ceramicservice, so I don't understand the need.

I mentioned this in the initial comment

> For the CeramicService work, I wanted to have a place in the app, where I can test things.
In other places, I would have add a playground route, but decided to just create a temporary "demo component" in the /storybook route.

I wanted to create a "dummy app" with some "Create" and "Load" buttons calling the ceramic service

src/resources/elements/primeDesignSystem/demos/demos.ts
hiaux0: Right, was lost from copy paste.

dkent600 (changes requested):
Can we use the existing demos.html file, and the aurelia support for enumerating "nav" routes (see here:  https://aurelia.io/docs/routing/configuration#basic-configuration)

hiaux0:
Keep enumaration approach for simplicty

PR #101, opened 2022-01-28
Rename, improve and organize deal related interfaces

paweljanicki:
## What was done
- Move deal registration data related interfaces out of the dealWizard directory to Deal entity. This was done because they don't only belong to the wizard, they will be used on homepage, deal dashboard, etc
- Create `IDealRegistrationData` interface that reflects current structure of the registration data, and should be the only one interface for the deal registration data. In a way it replaced the `IDealConfig`
- `IWizardState` now has a type variable Data, which specifies type of registrationData. So all component using `IWizardState` can pass a type for the registrationData they expect (deal wizard related components will pass `IDealRegistrationData`)


src/services/DealService.ts
```diff
@@ -215,14 +142,14 @@ export class DealService {
     }
   }
 
-  public async createRegistration(registration: IDealConfig): Promise<void> {
+  public async createRegistration(registration: IDealRegistrationData): Promise<void> {
```
dkent600, line 145: I think these two functions should be in the Deal entity.

dkent600 (approved):

src/services/WizardService.ts
```diff
@@ -2,10 +2,10 @@ import { autoinject } from "aurelia-framework";
 import { Router, RouterConfiguration, NavigationInstruction, RouterEvent } from "aurelia-router";
 import { EventAggregator } from "aurelia-event-aggregator";
 
-export interface IWizardState {
+export interface IWizardState<Data = any> {
   stages: Array<IWizardStage>;
   indexOfActive: number;
-  registrationData: IRegistrationData;
+  registrationData: Data;
```
hiaux0, line 8: @paweljanicki 
Could you provide a sample Interface for `Data` other than `IDealRegistrationData`?

I remember you mentioned, that `Data` can be anything.
Currently in the code, though, only `IDealRegistrationData` is ued for the `Data` generic here.
So I'm wondering about the future use case of the `Data` generic.

PR #103, opened 2022-01-28
Add a wizardManager that could be used for any kind of deal by configuring stages

paweljanicki:
The purpose of this PR is to demonstrate that one wizardManager could handle multiple wizard type and scenarios. For example, openProposal, partneredDeal, makingAnOffer, editing deals 

## What was done
- create generic `WizardManager` view-model
- configure stages based on wizardType
- Add routes for different wizardTypes, that use the same WizardManager component
- Pass wizardType into each stage, so they can perform different logic based on that
- Delete openProposalWizardManager, partneredDealWizardManager and makeOfferWizardManager
- Update route paths for each stage so they reflect what the stage is doing and not with stage in a row it is

You can test it here: https://prime-deals-dapp-git-single-wizard-manager-curvelabs.vercel.app/initiate/token-swap/open-proposal/proposal


src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,78 @@
...
+        break;
+    }
+
+    this.addWizardTypeToStages(wizardType);
+  }
+
+  private getWizardType(): WizardType {
+    // @TODO add logic
+
+    return WizardType.partneredDeal;
```
paweljanicki, line 66: The method should perform real logic and decide which WizardType it should be

src/dealWizard/dealWizardTypes.ts
```diff
@@ -1,5 +1,7 @@
 import { IWizardState } from "services/WizardService";
 
+export enum WizardType {"openProposal", "partneredDeal", "makeAnOffer"}
```
IonelLupu, line 3: enums can have values without quotes

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.configureStages();
+    this.wizardState = this.wizardService.registerWizard(this, this.stages, this.registrationData);
+  }
+
+  public onClick(index: number): void {
+    this.wizardService.goToStage(this, index);
+  }
+
+  private configureStages() {
+    const wizardType = this.getWizardType();
```
IonelLupu, line 42: can we have a call to discuss about this switch statement?

hiaux0:
> At the moment it is hardcoded to reflect "partneredDeal" wizard type.

It would be easier for me to see the benefit (even though I know there is), when there are multiple usages of this generic WizardManager.
Ie., I expected to see a negative diff for the openProposal and partneredDeal, to show how this WizardManager behaves.

If my request is out of scope, just react with 👎 and ignore

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,87 @@
...
+    name: "Primary DAO",
+    valid: false,
+    route: "stage3",
+    moduleId: PLATFORM.moduleName("./stages/primaryDaoStage/primaryDaoStage"),
+  }];
+  private registrationData = new DealRegistrationData();
+
+  constructor(public wizardService: WizardService) {}
+
+  activate(params: any, routeConfig: RouteConfig): void {
```
hiaux0, line 34: Could you add a type here? sth like

`StageRouterParam = { stageRoute : string }`.

Maybe also a comment, that `*stageRoute` is used in `app.ts`.
Though I'm not sure, if we should hard code file names here.
Maybe "used in wizard router configuration"

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    this.wizardState = this.wizardService.registerWizard(this, this.stages, indexOfActive, this.registrationData);
+  }
+
+  public onClick(index: number): void {
+    this.wizardService.goToStage(this, index);
+  }
+
+  private configureStages(wizardType: WizardType, parentRoutePath: string) {
+    switch (wizardType) {
```
hiaux0, line 63: Nit: I somehow find having `private stages = [...]` and then have this method, that the `stages` data is not cohesive enough for me.
So the idea is, to move `private stages ...`
1. into this method as `const stages = [...]`
2. Create a separate method `getBaseStages(): IWizardStage[]`
3. Maybe even a StagesService / StagesManager

Anyway, just writing down my thoughts as they come.
Fine with leaving as is.

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,87 @@
...
+        break;
+    }
+
+    this.stages = this.updateStagesWithFullRoutePath(parentRoutePath);
+  }
+
+  private updateStagesWithFullRoutePath(parentRoute) {
+    return this.stages.map(stage => ({
+      ...stage,
+      route: parentRoute.replace("*stageRoute", stage.route),
```
hiaux0, line 84:
```suggestion
     const STAGE_ROUTE_PARAMETER = "*stageRoute";
      //...
      route: parentRoute.replace("*stageRoute", stage.route),
```
Reason: Feels a bit fragile to have this hard coded here

hiaux0 (commented):
Looks good so far.
Glad that index thing was a minor fix

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
...
+  public wizardState: IWizardState<IDealRegistrationData>;
+  public stageMeta: IStageMeta;
+  public view: string;
+  public viewModel: string;
+  private stages: IWizardStage[] = [];
+  private registrationData = new DealRegistrationData();
+
+  constructor(public wizardService: WizardService) {}
+
+  activate(params: {[STAGE_ROUTE_PARAMETER]: string}, routeConfig: RouteConfig): void {
```
dkent600, line 47: I don't understand why the brackets `[STAGE_ROUTE_PARAMETER]`

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
+import { autoinject } from "aurelia-framework";
+import { PLATFORM } from "aurelia-pal";
+import { RouteConfig } from "aurelia-router";
+import { WizardService, IWizardState, IWizardStage } from "services/WizardService";
+import { DealRegistrationData, IDealRegistrationData } from "entities/DealRegistrationData";
+import { IStageMeta, WizardType, STAGE_ROUTE_PARAMETER } from "./dealWizardTypes";
+
+const proposalStage: IWizardStage = {
```
dkent600, line 8: I think would be better to scope these routes inside the class

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
...
+  name: "Partner DAO",
+  valid: false,
+  route: "partner-dao",
+  moduleId: PLATFORM.moduleName("./stages/partnerDaoStage/partnerDaoStage"),
+};
+
+@autoinject
+export class WizardManager {
+  public wizardState: IWizardState<IDealRegistrationData>;
+  public stageMeta: IStageMeta;
```
dkent600, line 39: can you please comment this and the other significant properties, as well as providing instructions on how to invoke a wizard?

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!this.wizardService.hasWizard(this)) {
+      this.stages = this.configureStages(wizardType, parentRoutePath);
+    }
+
+    this.stageMeta = {
+      wizardType,
+      wizardManager: this,
+    };
+
+    const indexOfActive = this.stages.findIndex(stage => stage.route.includes(stageRoute));
```
dkent600, line 63: would be nice to comment what you are setting up here

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private updateStagesWithFullRoutePath(stages: IWizardStage[], parentRoute) {
+    return stages.map(stage => ({
+      ...stage,
+      route: parentRoute.replace(`*${STAGE_ROUTE_PARAMETER}`, stage.route),
+    }));
+  }
+
+  private getOpenProposalStages(): IWizardStage[] {
```
dkent600, line 100: these can be properties instead of functions

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,116 @@
...
+  constructor(public wizardService: WizardService) {}
+
+  activate(params: {[STAGE_ROUTE_PARAMETER]: string}, routeConfig: RouteConfig): void {
+    if (!params[STAGE_ROUTE_PARAMETER]) return;
+
+    const stageRoute = params[STAGE_ROUTE_PARAMETER];
+    const wizardType = routeConfig.settings.wizardType;
+    const parentRoutePath = routeConfig.route as string;
+
+    if (!this.wizardService.hasWizard(this)) {
```
dkent600, line 54: I think it would be better if the service makes this check inside the function

dkent600 (changes requested):
I wonder if we can nest the folder "dealWizard" until a root folder called "wizards".  can be a separate PR.

src/dealWizard/wizardManager.ts
paweljanicki: It's a const - name of the route param
`export const STAGE_ROUTE_PARAMETER = "stageRoute";`

``` 
route: `/initiate/token-swap/partnered-deal/*${STAGE_ROUTE_PARAMETER}`
```

src/dealWizard/wizardManager.ts
paweljanicki: Sure, I don't see much difference, but I can move it

src/dealWizard/wizardManager.ts
paweljanicki: Alright

src/dealWizard/wizardManager.ts
paweljanicki: true

src/dealWizard/wizardManager.ts
paweljanicki: Good catch, actually it's not necessary to have this if statement at all. I will remove it

src/dealWizard/README.md
```diff
@@ -0,0 +1,54 @@
+# Deal Wizards
```
hiaux0, line 1: 🤩

hiaux0: Argh, but looking at this.
I'm wondering how future proof this is.
I doooo like, that we have it, but the underlying code might change fast, so this becomes outdated quickly. 🤔 

Overall, I think this still has lots of value, just because the wizard is touching multiple places, that makes 
"looking at the code" not trivial.

src/dealWizard/dealWizardTypes.ts
```diff
@@ -1,9 +1,18 @@
...
 export interface IBaseWizardStage {
   wizardManager: any;
   wizardState: IWizardState;
 
   activate;
   attached: () => void;
 }
+
+export interface IStageMeta {
+  wizardManager: any;
```
hiaux0, line 14:
```suggestion
export interface IStageMeta<T> {
  wizardManager: T;
```

Dunno if makes sense here.

src/dealWizard/wizardManager.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // a meta configuration passed to each stage component in the view
+  // for stage components to know which wizardManger they belong to and what wizardType it is
+  public stageMeta: IStageMeta;
+
+  // view of the currently active stage
+  public view: string;
+
+  // view model of the currently active stage
+  public viewModel: string;
```
hiaux0, line 20:
```suggestion
  public viewPath: string;

  // view model of the currently active stage
  public viewModelPath: string;
```

Self-explanatory? yay - nay?

hiaux0 (commented):
LGTM overall

dkent600 (approved):

PR #106, opened 2022-01-31
Added stage 2 of the registration wizard using the aurelia-validation…

IonelLupu:
## What was done
The stage 2 of the registration wizard was added. Used the aurelia-validation package for form validation.

Preview: https://prime-deals-dapp-git-feature-registration-wiza-20db25-curvelabs.vercel.app/initiate/token-swap/open-proposal/proposal-lead

The form validation is automatically done by using the (VERY IMPORTANT)  ` this.form.validateTrigger  = "changeOrBlur" `.

I also added a custom ValidationController renderer that can be used to automatically display errors in `pform-input` when the `pform-input` has only one element of type "input". I will use this in the next stages of the wizard or any other form.

It means there are two ways of displaying errors: 
1. you listen to the validation event, collected the errors and manually pass them to `pform-input` (used in this PR).
2. you just add the renderer to the validation form (`form.addRenderer(new PrimeRenderer)`) and let it add the errors on your `pform-inputs`. Again, make sure your `pfrom-inputs` have only ONE element.


paweljanicki:
Looks good to me! Let's go over the merge conflicts in a call. The new approach to validation that you introduced, after the merge conflicts are resolved, let's modify wizardService to handle that, and maybe other stage components if needed.

src/dealWizard/validation.ts
```diff
@@ -0,0 +1,26 @@
+import { ControllerValidateResult, Rule, ValidationController, ValidationRules } from "aurelia-validation";
+import { IProposalLead } from "../entities/Deal";
+
+export const proposalLeadValidationRules = ValidationRules
+  .ensure<IProposalLead, string>(proposalLead => proposalLead.address)
+  .satisfiesRule("isETHAddress")
```
hiaux0, line 6: Wdyt about creating an enum or const for `isETHAddress`?
Can then have some kind of map for all the custom validation rules

src/dealWizard/components/proposalLeadSection/proposalLeadSection.html
```diff
@@ -4,23 +4,39 @@
...
+        </pbutton>
       </pform-input>
 
+
       <pform-input
         class="wizardFormInput wizardFormInputNarrow"
-        label="Contact E-mail (optional)">
+        label="Contact E-mail (optional)"
+        validation-message.bind="errors.email"
+        validation-state.bind="errors.email ? 'error' : ''"
```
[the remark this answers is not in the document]
IonelLupu: I thought about that but the validation-state can also be `warning` or `validating` and not only `error`.

giladt:
Is it possible to resolve error messages after the textbox is cleared or a valid value is entered?
I think it might provide a better UX.
input errors

IonelLupu:
> Is it possible to resolve error messages after the textbox is cleared or a valid value is entered? I think it might provide a better UX. input errors
 @giladt 
@hiaux0 was also mentioning this. I tried to do so but it seems is a lot more complicated that I thought because Aurelia has a weird way of doing validation. We need to investigate this and implement it on every form we have for better UX.

dkent600 (changes requested):
No requests, just to get this out of my requested-review list

giladt (changes requested):
Having a look at the preview, I've noticed 2 issues you might like to address:
1. EMail validation- entering an email with the format `xxx@yyy` is valid.
2. Stage one (if not already on your agenda), can you please apply these changes there too, so that the validations will be resolved upon input changes?

src/dealWizard/components/proposalLeadSection/proposalLeadSection.html
```diff
@@ -4,24 +4,41 @@
       <h2 class="heading heading3 title">Proposal Lead</h2>
       <p class="subtext">
         Provide the Ethereum address for the proposal lead who will gain admin rights of the deal.
-        <br />
-        <br />
-        Proposal leads will be responsible for editing the deal during negotiation, initiating the funding phase once a deal is approved, and executing the token swap. You don’t need to be part of a DAO to be the lead proposal.
+        <br/>
```
dkent600, line 7: Ii think this should be done with a class + margin or padding, rather then brs

src/dealWizard/components/proposalLeadSection/proposalLeadSection.ts
```diff
@@ -1,9 +1,38 @@
...
+  ethAddress?: string;
+
+  private accountSubscription: Subscription;
+
+  constructor(
+    private eventAggregator: EventAggregator,
+    private ethereumService: EthereumService,
+  ) {
+    this.ethAddress = this.ethereumService.defaultAccountAddress;
+    this.accountSubscription = this.eventAggregator.subscribe("Network.Changed.Account", address => {
```
dkent600, line 25: should be located in attached, to match the detached

src/dealWizard/components/proposalLeadSection/proposalLeadSection.ts
```diff
@@ -1,9 +1,38 @@
...
+      this.ethAddress = address;
+    });
+
+  }
+
+  async connectToWallet() {
+    await this.ethereumService.connect();
```
dkent600, line 32: no need for await here.  Doesn't behave as you might expect, anyways

src/dealWizard/openProposalWizard/openProposalProposalLeadStage/openProposalProposalLeadStage.ts
```diff
@@ -1,33 +1,39 @@
...
 @autoinject
 export class OpenProposalProposalLeadStage implements IBaseWizardStage {
   public wizardManager: any;
   public wizardState: IWizardState<IDealRegistrationTokenSwap>;
   public errors: WizardErrors<IProposalLead> = {};
+  public form: ValidationController;
 
-  constructor(public wizardService: WizardService) {}
+  constructor(public wizardService: WizardService, validationFactory: ValidationControllerFactory) {
+    this.form = validationFactory.createForCurrentScope();
```
dkent600, line 18: can we make this automatically happen for all stages by moving it into the WizardService?

src/dealWizard/openProposalWizard/openProposalProposalLeadStage/openProposalProposalLeadStage.ts
```diff
@@ -1,33 +1,39 @@
...
   }
 
   attached(): void {
     this.wizardState = this.wizardService.getWizardState(this.wizardManager);
-    this.wizardService.registerStageValidateFunction(this.wizardManager, this.validate.bind(this));
-  }
-
-  validate(): boolean {
-    this.errors = {};
+    this.form.addObject(this.wizardState.registrationData.proposalLead, proposalLeadValidationRules.rules);
```
dkent600, line 29: I would hate to have to jump though all these hoops for every state validation function.  Is possible to make the hoops reusable somehow?

src/dealWizard/components/proposalLeadSection/proposalLeadSection.ts
```diff
@@ -1,9 +1,38 @@
 import { bindable } from "aurelia-typed-observable-plugin";
 import { IProposalLead } from "entities/DealRegistrationTokenSwap";
 import { WizardErrors } from "services/WizardService";
+import "./proposalLeadSection.scss";
```
dkent600, line 4: Why have proposalLeadSection if only used once?

dkent600 (changes requested):
can we just have one proposalLeadStage component?  Our pattern is to use the wizardType to customize each tage according to the context.  SHould go under the stages folder.  Check with @paweljanicki on this.  we should delete the open and partnered stages stuff, which is deprecated.

src/dealWizard/components/proposalLeadSection/proposalLeadSection.ts
dkent600: ok, this can be removed once you have just a single stage component

IonelLupu: This component is actually used three times throughout the app. Or what do you mean?
image

src/dealWizard/openProposalWizard/openProposalProposalLeadStage/openProposalProposalLeadStage.ts
IonelLupu: what do you mean?

src/dealWizard/components/proposalLeadSection/proposalLeadSection.ts
IonelLupu: It's good practice to await promises usually. (and my IDE complains about it. I just wanted to get rid of the warning actually 🤣 )

What do you mean about "Doesn't behave as you might expect" ?

src/dealWizard/openProposalWizard/openProposalProposalLeadStage/openProposalProposalLeadStage.ts
IonelLupu: I moved this in the proposalLeadSection component

src/dealWizard/components/proposalLeadSection/proposalLeadSection.html
IonelLupu: I say those while editing (were there before) and I thought about the same thing. I added design system spacing classes and used those instead

IonelLupu:
@giladt proper email validation and stage 1 validation added

dkent600 (changes requested):
Just resetting the review request

src/main.ts
```diff
@@ -17,13 +17,15 @@ import { TokenService } from "services/TokenService";
 import { CeramicServiceMock } from "services/CeramicServiceMock";
 import { DealTokenSwap } from "entities/DealTokenSwap";
 import { IDataSourceDeals } from "services/DataSourceDealsTypes";
+import "./validation";
```
dkent600, line 20: do we need this line?

src/resources/elements/primeDesignSystem/pform-input/pform-input.ts
```diff
@@ -15,46 +15,49 @@ export class PFormInput {
...
 
+  validationStateChanged(newValue?: ValidationState) {
+    if (this.inputReference) {
+      this.inputReference.validationState = newValue;
+    }
+  }
+
   private limitInputCharacterLength() {
-    const inputValueObserver = this.bindingEngine.propertyObserver(this.input, "value");
+    const inputValueObserver = this.bindingEngine.propertyObserver(this.inputReference, "value");
```
dkent600, line 53: FYI we have AureliaHelperService for doing this.  no worries.

src/resources/elements/primeDesignSystem/pform-input/pform-input.ts
```diff
@@ -15,46 +15,49 @@ export class PFormInput {
   @bindable.string helperMessage = "";
   @bindable validationMessage = "";
   @bindable.string @observable validationState?: ValidationState;
+  @bindable inputReference;
```
dkent600, line 18: ` @bindable.string` is alreadly observable.  No need for `@observable`.

dkent600: would be good if you can comment what inputReference is about.

src/validation.ts
```diff
@@ -0,0 +1,20 @@
+import { ValidationRules } from "aurelia-validation";
```
dkent600, line 1: "src" is not an appropriate location.   I recommend a `ValidationService.ts` in services.

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "sourceMap": true,
     "target": "es2019",
     "lib": [
-      "es2019",
+      "ESNext",
```
dkent600, line 15: why?

dkent600 (changes requested):

tsconfig.json
IonelLupu: I added this library in order to have access to the latest JS features (like String.replaceAll() )

src/validation.ts
IonelLupu: it's not really a service. Is just a file with custom validation rules. Is there a place for such things?

src/main.ts
IonelLupu: Yes. Aurelia will read this file and register all the custom validation rules from it

src/validation.ts
dkent600: I would put it where I suggested.  I think calling it a service is ok.

dkent600 (approved):

PR #107, opened 2022-02-01
Added the `value` attribute for pselect

IonelLupu:
## What was done
The `<pselect>` can now accept the `value` attribute

The `value` property was added in order to programmatically set a value for the select input. This property is also available for all the other input elements like `pinput-text`,`pinput-numeric`,`ptextarea`, `ptoggle` etc. 

It can be used to set a default value, but most importantly, it allows you to get the value out of the `pselect`


src/resources/elements/primeDesignSystem/pselect/pselect.ts
```diff
@@ -34,36 +34,33 @@ export interface IPSelectItemConfig {
...
-    if (this.data) {
-      selectConfig.data = [
+      data: [
         {text: this.placeholder, placeholder: true},
         ...this.data,
-      ];
-    }
-
-    this.select = new SlimSelect(selectConfig);
+      ],
```
hiaux0, line 60: How is this related to the `value` attribute addition?

src/resources/elements/primeDesignSystem/pselect/pselect.ts
```diff
@@ -34,36 +34,33 @@ export interface IPSelectItemConfig {
...
+      onChange: info => this.value = info instanceof Array ? info.map(item => item.value) : info.value,
```
hiaux0, line 61:
```suggestion
      onChange: info => this.value = Array.isArray(info) ? info.map(item => item.value) : info.value,
```

Nit: Maybe just preference.

src/resources/elements/primeDesignSystem/demos/pselectDemo.html
```diff
@@ -1,7 +1,10 @@
 <template>
   <section class="demo">
     <h1>Simple Dropdown</h1>
-    <pselect data.bind="simpleData"></pselect>
+    --${test}--
+    <pselect data.bind="simpleData" value.bind="test"></pselect>
```
hiaux0, line 5: Obsolete dev code?

Could we have a demo, where an array of `value`s is set?

dkent600: what is `  --${test}--` for?

dkent600 (changes requested):
@IonelLupu Can you describe in the description why we care about `value`?  If this is for a default, maybe we can call it `default` or something like that?

src/resources/elements/primeDesignSystem/pselect/pselect.ts
IonelLupu: didn't know about. love it

src/resources/elements/primeDesignSystem/pselect/pselect.ts
IonelLupu: I couldn't type the `selectConfig` object in order to get intellisense so I decided to instantiate SlimSelect using the object directly. This way I prevented possible errors when adding SlimSelect properties.

IonelLupu:
@dkent600 I updated the description of this PR
@hiaux0  I also added the `multiple` feature for the `pselect`

dkent600 (approved):
I see this isn't just about adding `value`, but also about multiple.  Should probably mention that in the description

IonelLupu:
@dkent600 updated the description

PR #114, opened 2022-02-02
Fix previous button on first stage of wizard not redirecting to do you have a partner screen

paweljanicki:
- Update `registerWizard` method to accept cancelRoute and previousRoute
- Use cancelRoute and previousRoute to navigate in the cancel and previous methods of wizardService

### DIFF
--- a/src/services/WizardService.ts
+++ b/src/services/WizardService.ts
@@ -6,6 +6,8 @@ export interface IWizardState<Data = any> {
   stages: Array<IWizardStage>;
   indexOfActive: number;
   registrationData: Data;
+  cancelRoute: string;
+  previousRoute: string;
 }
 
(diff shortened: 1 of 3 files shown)


dkent600 (approved):

src/services/WizardService.ts
```diff
@@ -88,7 +88,7 @@ export class WizardService {
     if (indexOfActive > 0) {
       this.goToStage(wizardManager, indexOfActive - 1);
     } else {
-      this.router.parent.navigate("initiate/token-swap");
+      this.router.navigate("initiate/token-swap");
```
dkent600, line 91: this wizard should be wizard-agnostic.  Should not know about "token-swap".

dkent600 (changes requested):

dkent600 (approved):

PR #118, opened 2022-02-04
Add e2e tests to open proposal and partnered deal wizards, for stepper and stage one

paweljanicki:
## What was done
- New e2e tests for open proposal and partnered deal, covering stepper navigation and proposal stage


cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
+import { Then, When } from "@badeball/cypress-cucumber-preprocessor/methods";
+
+Then("I am presented the option to choose a partner", () => {
+  cy.url().should("match", /(initiate\/token-swap$)/)
+})
```
hiaux0, line 5: This only checks the url, but should also verify "the option to choose a partner"

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+
+When("I go to previous step", () => {
+  cy.get("[data-test='wizard-previous-button']").click();
+})
```
hiaux0, line 9: I'm thinking, that we could also verify the url here

1. get current url
2. click button
3. check, that url changed

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+
+When("I try to proceed to next step", () => {
+  cy.get("[data-test='wizard-proceed-button']").click();
+})
```
hiaux0, line 13: The same as previous with url verifying

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+When("I try to navigate to partner dao stage via stepper", () => {
+  cy.get("[data-test='wizard-manager-stepper']").within(() => {
+    cy.contains("[data-test='pstepper-step']", "Partner DAO").click();
+  });
+})
+
+When("I fill in proposal title correctly", () => {
+  cy.get("[data-test='proposal-title-field']").within(() => {
+    cy.get('input').type('Test proposal')
+  })
```
hiaux0, line 42: You write "correctly", but don't verify correctness here, So either
1. remove "correctly" in description
2. add correctness

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+})
+
+When("I fill in proposal summary with text that is too short", () => {
+  cy.get("[data-test='proposal-summary-field']").within(() => {
+    cy.get('textarea').type('asd')
+  })
+})
```
hiaux0, line 49: > "too short"

Same as "correctly"

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+
+When("I fill in proposal description with text that meets requirements", () => {
+  cy.get("[data-test='proposal-description-field']").within(() => {
+    cy.get('textarea').type('Test description')
+  })
+})
+
+Then("I am presented with errors proposal stage required errors", () => {
```
hiaux0, line 69: This reads off a bit.
Did you mean sth like this?:

```suggestion
Then("I am presented with proposal stage required errors", () => {
```

cypress/integration/common/deal-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+  })
+  cy.get("[data-test='proposal-summary-field']").within(() => {
+    cy.get('.errorMessage').should('contain.text', 'Required Input')
+  })
+  cy.get("[data-test='proposal-description-field']").within(() => {
+    cy.get('.errorMessage').should('contain.text', 'Required Input')
+  })
+})
+
+Then("I am presented with too short input errors for proposal summary and description", () => {
```
hiaux0, line 81: Future: we may want to split up a separate step for each field.
For now it's fine like this, so we can observe how this (=complexity of validation) behaves

cypress/integration/features/deals/create-open-proposal-wizard.feature
```diff
@@ -0,0 +1,46 @@
...
+
+  Scenario: I proceed from proposal stage without filling in required fields
+    When I try to proceed to next step
+    Then I am presented with Open Proposal proposal stage
+    And I am presented with errors proposal stage required errors
+
+  Scenario: I proceed from proposal stage after filling required fields with too short summary and description
+    When I fill in proposal title correctly
+    And I fill in proposal summary with text that is too short
+    And I fill in proposal description with text that is too short
```
hiaux0, line 18: Re: reusable-step-definitions
https://github.com/TheBrainFamily/cypress-cucumber-preprocessor#reusable-step-definitions
```
    And I fill in proposal  with text that is too short
    And I fill in proposal description with text that is too short
```

In the test file you can reduce duplication

Note: Here I'm using regex matches, which is also supported 
https://github.com/cucumber/cucumber-expressions#custom-parameter-types (--> note the `regex` entry)
Here is a list of predefined "parameters": https://github.com/cucumber/cucumber-expressions#parameter-types

```diff
- When("I fill in proposal summary with text that is too short", () => {
+ When(/^I fill in (.*) with text that is too short$/, (fieldType) => {
  // Note: Because in description we use "proposal summary", but the selector is `proposal-summary-field`
  // We could either
  // 1. Define a separate map
  // 2. try to convert the description string to the selector string
```

cypress/integration/features/deals/create-open-proposal-wizard.feature
```diff
@@ -0,0 +1,46 @@
...
+    When I try to navigate to primary dao stage via stepper
+    Then I am presented with Open Proposal primary dao stage
+    When I try to navigate to proposal lead stage via stepper
+    Then I am presented with Open Proposal proposal lead stage
+    When I try to navigate to proposal stage via stepper
+    Then I am presented with Open Proposal proposal stage
+    When I try to navigate to primary dao stage via stepper
+    Then I am presented with Open Proposal primary dao stage
+    When I try to navigate to proposal stage via stepper
+    When I am presented with Open Proposal proposal stage
```
hiaux0, line 46: First thoughts:
- too long
- a lot of repeation

I'll submit a PR later on how this could be handled

cypress/integration/features/deals/create-partnered-deal-wizard.feature
```diff
@@ -0,0 +1,53 @@
...
+    When I fill in proposal title correctly
+    And I fill in proposal summary with text that meets requirements
+    And I fill in proposal description with text that meets requirements
+    And I try to proceed to next step
+    Then I am presented with Partnered Deal proposal lead stage
+
+  Scenario: Wizard stepper contains correct stages
+    Then I can see stages correct for Partnered Deal
+
+  Scenario: I can navigate the stages freely using stepper
```
hiaux0, line 33: This somehow sounds like it should be a dedicated component test of the stepper, as the underlying logic does not really require the wizard

cypress/integration/features/deals/create-partnered-deal-wizard.feature
```diff
@@ -0,0 +1,53 @@
+Feature: Create Partnered Deal
+  Background:
+    Given I navigate to create partnered deal wizard
+
+  Scenario: Go back to deal type select
+    Given I am presented with Partnered Deal proposal stage
+    When I go to previous step
+    Then I am presented the option to choose a partner
+
+  Scenario: I proceed from proposal stage without filling in required fields
```
hiaux0, line 10:
```suggestion
  Scenario: Proceed from proposal stage without filling in required fields
```

Maybe just preference, but Scenarios should be formulated in a general sense.
Exception: 
- Admin is not allowed to xyz
- others??

hiaux0 (commented):
## Code

- "too short", "correctly", "meet requirements"
    - comments I made apply to all. I didn't comment on all places

- overall looks good

cypress/integration/common/deal-wizard.ts
paweljanicki: There is already a scenario that covers that in choose-deal-type.feature:

```
  Scenario: Verify correct Token Swap types
    Given I select Token Swap
    Then I can see Open Proposal and Partnered Deal
```

Is there a need to double that?

cypress/integration/common/deal-wizard.ts
paweljanicki: We could check if it changed, like:
```
When("I go to previous step", () => {
 const initialUrl
  cy.get("[data-test='wizard-previous-button']").click();
const finalUrl

test initialUrl === finalUrl
})
```

Is that what you mean?

Scenarions that use this step, check the url in the following step:

```
  Scenario: Go back to deal type select
    Given I am presented with Open Proposal proposal stage
    When I go to previous step
    Then I am presented the option to choose a partner
```

I wanted to make that step generic

cypress/integration/common/deal-wizard.ts
paweljanicki: What I meant is that it's filled in correctly according to the business logic. I'm not sure if that implies checking for error messages in the same step but I don't know I'm not really into linguistics :D  
I can change it to "I fill in proposal title"

cypress/integration/features/deals/create-open-proposal-wizard.feature
paweljanicki: I can look into that

cypress/integration/features/deals/create-open-proposal-wizard.feature
paweljanicki: That's a real life user flow. I think it should be tested, looking forward to your PR

cypress/integration/features/deals/create-partnered-deal-wizard.feature
paweljanicki: Not really, because it's the wizard that is handling the navigation.
```
  public goToStage(wizardManager: any, index: number): void {
    const wizardState = this.getWizardState(wizardManager);
    wizardState.indexOfActive = index;

    const params = {
      ...this.router.currentInstruction.params,
      [STAGE_ROUTE_PARAMETER]: wizardState.stages[index].route,
    };

    this.router.navigateToRoute(
      this.router.currentInstruction.config.name,
      params,
    );
  }
```
If something breaks here (maybe someone decides to add validation here) it will break test.

paweljanicki: I think we should have:
- e2e test for pStepper component
- e2e test for using a the stepper in all Wizards
- unit tests for `goToStage` method and other methods in wizardService

cypress/integration/common/deal-wizard.ts
paweljanicki: I don't know how to check if url changed, because in cypress we cannot variables in the way we do in normal JS. If you know the solution please suggest it

cypress/integration/common/deal-wizard.ts
hiaux0: In that case what is the difference between
1. Then I am presented the option to choose a partner
2. Then I can see Open Proposal and Partnered Deal
?

All I'm trying to say is, that based on what I read "I am presented the option to choose a partner", I expected to test to verify for the option, because it is expecially the `Then` step.

I do see a challenge to differentiate on what to put into Scenario, and what to put into Steps,
and can't really give recommendations at the top of my head.

If you feel, the Steps and Scenarios are fine as is, then fine with me.
We can refactor later if ever confusion comes up.

cypress/integration/common/deal-wizard.ts
hiaux0: > generic

True.
Though I don't see verifying "url changed" against "being generic".

If I read this Scenario from a Engineer perspective
```feature
Scenario: Go back to deal type select 
  Given I am presented with Open Proposal proposal stage 
  When I go to previous step 
  Then I am presented the option to choose a partner
```

Then in the `When` step, is where I would have put the "generic url test".

---
Also, that wouldn't be a problem, if multiple steps test for the url.
If Steps are generic they should cover a bit rather than a bit less imo.

hiaux0: > suggestions

Can you check if this works
```ts
cy.url().then(url => {
  const oldUrl = url;
  // do click
  cy.get().click()

  // verify url change
  cy.url().should('equal', oldUrl);
})
```

cypress/integration/common/deal-wizard.ts
hiaux0: My point was, that the step did it, but didn't verify, that "it is too short"

cypress/integration/common/deal-wizard.ts
hiaux0: If you just use this step in the Scenario

```feature
  Scenario: I proceed from proposal stage after filling required fields correctly
    When I fill in proposal title correctly
    And I fill in proposal summary with text that meets requirements
    And I fill in proposal description with text that meets requirements
    And I try to proceed to next step
    Then I am presented with Open Proposal proposal lead stage
```

Then I would after that `When` step, that UI has no error states.
We can argue that "And I try to proceed to the next step" handles the verification.

My intuition here is to treat the Steps, like actually going through the UI as a user.
If I fill out a field "correctly", I also as a side effect check, that the field is not red.

---
Anyway, fine with leaving as is, and add more to it later.

cypress/integration/features/deals/create-open-proposal-wizard.feature
hiaux0: Given things on my plate currently, I won't submit something in the near future, so leaving like this is okay.

cypress/integration/features/deals/create-partnered-deal-wizard.feature
hiaux0: Ah okay, I somehow thought stepper will always change url, but it doesn't have to.

That last suggestion makes sense 👍

hiaux0 (approved):
Don't want to be a blocker on this.
All comments are more of suggestions, and thus can be left out for this PR.

Given our coverage I'm a fan of getting tests in as quickly as possible, then iterate on them.
Many tests > few detailed ones

cypress/integration/common/deal-wizard.ts
paweljanicki: I am presented the option to choose a partner - This is just checking the URL
I can see Open Proposal and Partnered Deal - This is checking the UI

I'm not sure what's the correct way to handle this ;) 
It's a good idea to leave it as it is and dedicate some time for refactoring later on

cypress/integration/common/deal-wizard.ts
paweljanicki: For sure, I completely agree that we should check if url changed, sorry for misunderstanding. 

Thanks for the code snippet, it works!

cypress/integration/common/deal-wizard.ts
paweljanicki: This one is intentionally just about clicking the button, that is trying to proceed. Because we don't want to proceed if the inputs are not valid. So this steep is used in combination with other steps that check what happened

cypress/integration/common/deal-wizard.ts
paweljanicki: I think it has to stay like that for now, but later on if there is a better way lets improve it.
So at the moment "I try to proceed to next step" is only about clicking the proceed button.
The "I am presented with Open Proposal proposal lead stage" step is checking if the user is presented with the next stage and that's how we assume that the proposal stage passed the validation. 
That's true we could and probably should check if there are actual errors display, but we could only do that after the "proceed" button is clicked as that is when the validation happens (at least at the moment)

cypress/integration/common/deal-wizard.ts
paweljanicki: Ok, fair point

cypress/integration/features/deals/create-partnered-deal-wizard.feature
paweljanicki: Cool

cypress/integration/features/deals/create-partnered-deal-wizard.feature
paweljanicki: Makes sense

PR #125, opened 2022-02-06
Feature: added "in progress" on the deal page

IonelLupu:
## What was done
Linked the home page buttons to redirect to the correct pages

### DIFF
--- a/src/resources/elements/primeDesignSystem/_styles.scss
+++ b/src/resources/elements/primeDesignSystem/_styles.scss
@@ -1,2 +1,6 @@
 @import "./pinput";
 @import "./typography";
+
+.text-center {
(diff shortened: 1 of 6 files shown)


src/resources/elements/primeDesignSystem/_styles.scss
```diff
@@ -1,2 +1,6 @@
 @import "./pinput";
 @import "./typography";
+
+.text-center {
```
dkent600, line 4: This doesn't always work to center text.  And, is an example of the kind of naming that I want to avoid.

dkent600 (changes requested):

dkent600 (approved):

PR #133, opened 2022-02-08
implement pcountdown-circular component

dkent600:
## What was done
Implemented pcountdown-component.  Is designed to provide basic functionality that we can leverage in popup notifications, which will be done in a separate PR.

Demos:  https://prime-deals-dapp-git-pcountdown-circular-curvelabs.vercel.app/storybook/pcountdown-circular

demo

#### After


src/resources/elements/primeDesignSystem/_colors.scss
```diff
@@ -5,13 +5,15 @@ $Secondary02: #a258a7; // Accent color
 $Secondary03: #fff619; // Accent color
 $Secondary04: #ffce3d; // Accent color
 $Secondary05: #64557d; // Accent color
+$Purple-100: #64557d; // synonym for Secondary05
```
IonelLupu, line 8: why were these synonym added? I noticed the Figma design using these names but maybe they are old names. Can you confirm with Jaap if that's the case and use the "Secondary05" instead? If not, the best thing we can do is to assign the "Secondary05" variable to "$Purple-100": `$Purple-100: $Secondary05`

src/resources/elements/primeDesignSystem/demos/pCountdownCircularDemo.html
```diff
@@ -0,0 +1,74 @@
...
+    .ticked {
+      display: inline-block;
+    }
+
+    .clickable {
+      cursor: pointer;
+    }
+  </style>
+  <h1>Countdown from 10 seconds, alert on stopped</h1>
+  <div class="circularDemoContent">
```
IonelLupu, line 28: Not sure if it's a priority right now but the circular border looks jagged on Chrome when it's running. Should we try to fix it?

Also, not sure if it's a priority but should we animate the "ticks" between numbers? Right now it jumps from a number to another. It doesn't look that good when having a 3 second countdown for example.

IonelLupu (changes requested):

src/resources/elements/primeDesignSystem/_colors.scss
dkent600: Removed the synonyms because I think is best not to star tusing them yet.

src/resources/elements/primeDesignSystem/demos/pCountdownCircularDemo.html
dkent600: I would like to add animation, and have spent time on it, but have not been able to figure out how to do it, for either text or the curvy line.  Since it is not required in the mock, I don't think we should spend more time on it ATM.

IonelLupu (approved):

PR #135, opened 2022-02-09
Refactor/fetch homepage featured deals from deals service

giladt:
## What was done
Removed old Mock from `home.ts` and implemented it in the current ceramic mock.
Fixed `dealSummary` to implement changes in the mock
Fixed the `timeLeft` statuses to have the correct colors for each status.
Fixed `DealTokenSwap` to implement status in the deal object.
Added `DealStatus` enum in `IDealTypes.ts`.
Fixed navigation route to the correct deal dashboard route.

### DIFF
--- a/src/entities/DealTokenSwap.ts
+++ b/src/entities/DealTokenSwap.ts
@@ -77,6 +77,16 @@ export class DealTokenSwap implements IDeal {
     }
   }
 
+  /* ++++ TEMPORARY UNTIL STATUS LOGIC IS SORTED OUT ++++ */
+  private statuses = Object.values(DealStatus);
+  private shuffleArray(array): void {
+    for (let i = array.length - 1; i > 0; i--) {
+      const j = Math.floor(Math.random() * (i + 1));
+      [array[i], array[j]] = [array[j], array[i]];
+    }
+  }
+  /* ++++ ------------------------------------------ ++++ */
+
   private async hydrate(): Promise<void> {
     // eslint-disable-next-line no-empty
     try {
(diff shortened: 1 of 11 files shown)


src/entities/DealTokenSwap.ts
```diff
@@ -84,6 +84,7 @@ export class DealTokenSwap implements IDeal {
       this.registrationData = await this.dataSourceDeals.get<IDealRegistrationTokenSwap>(this.rootData.registration);
       const discussionsMap = await this.dataSourceDeals.get<Record<string, string> | undefined>(this.rootData.discussions);
       this.clauseDiscussions = new Map(Object.entries(discussionsMap ?? {}));
+      this.status = this.rootData.status;
```
dkent600, line 87: fake "status" should not be coming from Ceramic.  I think we need this coming from elsewhere for Deal entity stuff.

src/resources/elements/timeLeft/timeLeft.scss
```diff
@@ -7,18 +7,29 @@ $comingUpColor: $Orange;
 $comingUpBackground: #3e3038;
 $farOffColor: #cecece;
 $farOffBackground: #30343e;
-$liveColor: $Green;
+
+$completedColor: #18bab0;
```
dkent600, line 11: TO DO: These need to come from the color palette

src/entities/DealTokenSwap.ts
dkent600: Ideally we are deriving status by whatever logic it is supposed to be derived from

dkent600: How about for now we just have each Deal return a random value for status.

giladt: I've removed the ceramic implementation and added a (temporary) random status distributer in the DealTokenSwap entity.

src/resources/elements/timeLeft/timeLeft.scss
giladt: Moved the variables to the design system colors file. This still might change once @v0ort will update the palette in the Figma file.

dkent600 (approved):

PR #169, opened 2022-02-14
using colors from the DS palette

giladt:
## What was done
The colors in the `timeLeft` element are now taken from the design-system color palette.


giladt:
Pending @v0ort confirmation for the dark orange color naming.

dkent600 (approved):

PR #171, opened 2022-02-15
Added the stage 5 of the registration wizard

IonelLupu:
## What was done
Added the stage 5 of the registration wizard

Preview:  https://prime-deals-dapp-git-features-stage-5-registra-4d05f3-curvelabs.vercel.app/initiate/token-swap/partnered-deal/token-details
image


src/resources/elements/primeDesignSystem/pform-input/pform-input.html
```diff
@@ -26,7 +26,7 @@
   <!-- Validation message section -->
   <p
     class="feedbackMessage ${ validationStateExists(validationState) ? 'feedbackMessageVisible': '' } ${validationState + 'Message'}">
-    ${validationMessage}
+    ${!inputReference.disabled ? validationMessage : ''}
     <slot name="validation"></slot>
   </p>
```
hiaux0, line 31: Does it make sense to `if.bind` the `<p>` element here?
If I understand correctly, in the `disabled` case, we don't show any validation message..
..except for when `slot="validation"` is used?

I guess, I'm just a bit confused about the intended behavior of `pform-input` in case of
- validation message and
- disabled

Usage of `slot="validation"` example: (in the demo page)

```html
  <pform-input label="Name" validation-state.bind="field3">
    <pinput-text value.bind="field3"></pinput-text>

    <div slot="validation">       // <--- HERE
      <b style="color: orange"><i class="far fa-trash-alt"></i> This is my Custom validation </b> message
    </div>
  </pform-input
```

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
```diff
@@ -10,33 +10,39 @@ export class PRangeSlider {
...
-  @computedFrom("value")
-  get percentage() {
-    this.updateCssVariables();
-    return this.value ?? 0;
-  }
-
   @computedFrom("value", "maxValue")
   get leftValue() {
-    return this.percentageToAbsoluteValue(this.value);
+    return this.left = this.percentageToAbsoluteValue(this.value);
```
hiaux0, line 22: Is there another way to update `this.left`?
I'm feeling a bit uncomfortable with this side effect in this getter here.

src/services/TokenService.ts
```diff
@@ -257,7 +257,7 @@ export class TokenService {
 
       return axios.get(uri)
         .then((response) => {
-          tokenInfo.price = response.data.market_data.current_price.usd ?? 0;
+          // tokenInfo.price = response.data.market_data.current_price.usd ?? 0;
```
hiaux0, line 260: Needed or dev remnant?

src/wizards/tokenSwapDealWizard/components/stageButtons/stageButtons.ts
```diff
@@ -8,6 +8,13 @@ import "./stageButtons.scss";
...
   @bindable wizardManager: any;
   @bindable showSubmit = false;
+  validating = false;
 
-  constructor(public wizardService: WizardService) {}
+  constructor(public wizardService: WizardService) {
+  }
+
+  async proceed() {
+    this.validating = true;
```
hiaux0, line 17:
```suggestion
    this.isLoading= true;
```
Because, to me, it seems like you are only controlling the `is-loading` in the `<pbutton>` here

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
```diff
@@ -0,0 +1,158 @@
+<template>
+  <pcard if.bind="viewMode === 'edit'">
+    <pform-input
+      helper-message.bind="tokenInfoLoading ? 'Searching token address details...': ''"
+      input-reference.bind="addressInput"
```
hiaux0, line 5: Nit: 
(Best in another PR) (Fine with ignoring this)
The word "reference" is a bit ambiguous in the context of an Aurelia View.
Because one uses 
`<my-ele ref="xyz">`
more than
`<my-ele view-model.ref="abc">`

--> I'd suggest to use "reference" when we refer to an HTMLElement.
In other cases, we make it more explicit like:
```suggestion
      input-view-model.bind="addressInputViewModel"
```

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
```diff
@@ -0,0 +1,158 @@
...
+    >
+      <pinput-text
+        input.trigger="getTokenInfo(token.address)"
+        placeholder="0x..."
+        validation-state.bind="tokenInfoLoading? 'validating': addressInput.validationState !== 'validating' ? addressInput.validationState : undefined"
+        value.bind="token.address & validate"
+        view-model.ref="addressInput"
+      ></pinput-text>
+    </pform-input>
+    <div if.bind="infoSearched">
```
hiaux0, line 17:
```suggestion
    <div if.bind="showTokenDetails">
```
For me
`infoSearched` - is what the code is doing
`showTokenDetails` - what the variable is about

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
```diff
@@ -0,0 +1,158 @@
...
+            <p>Decimals</p>
+            <p if.bind="token.decimals">${token.decimals}</p>
+            <pinput-numeric if.bind="!token.decimals" value.bind="token.decimals"></pinput-numeric>
+          </td>
+          <td>
+            <p>Logo <img height="24" src.bind="token.logoURI"></p>
+            <pinput-text value.bind="token.logoURI"></pinput-text>
+          </td>
+        </tr>
+      </table>
```
hiaux0, line 42: Just random related thought:
Wondering what is easier to maintain `<table>` or `display: grid`

hiaux0 (commented):
## Code
- Looks good overall.
- Places where I've started with "Nit:" can be ignored

## Fncality
- not tested

## Tests
What's the status of e2e tests?

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
```diff
@@ -0,0 +1,122 @@
...
+  ValidationRules,
+} from "aurelia-validation";
+import { PrimeRenderer } from "../../../../resources/elements/primeDesignSystem/validation/primeRenderer";
+import { Validation } from "../../../../services/ValidationService";
+import { Utils } from "../../../../services/utils";
+
+@autoinject
+export class TokenDetails {
+  @bindable token: IToken;
+  @bindable.string() viewMode: "edit" | "view" = "edit";
```
hiaux0, line 20: Is `viewMode` used as a bindable?
Couldn't find it being used in `tokenDetailsStage.html` in `<token-details>`

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
```diff
@@ -0,0 +1,122 @@
...
+  ) {
+
+    this.form = this.validationControllerFactory.createForCurrentScope();
+    this.form.validateTrigger = validateTrigger.change;
+    this.form.addRenderer(new PrimeRenderer);
+  }
+
+  attached() {
+    ValidationRules.customRule(
+      "isValidIERC20Address",
```
dkent600, line 41: Maybe we can add this to the ValidationService so is reusable?  No big deal if is not anticipated to be needed again

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -0,0 +1,78 @@
...
+export class TokenDetailsStage {
+  wizardManager: any;
+  wizardState: IWizardState<IDealRegistrationTokenSwap>;
+  isOpenDealWizard = false;
+  form: ValidationController;
+
+  tokenDetailsForms: ValidationController[] = [];
+
+  constructor(
+    public wizardService: WizardService,
```
dkent600, line 18: why public?

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -0,0 +1,78 @@
...
+  ) {
+  }
+
+  @computedFrom("isOpenDealWizard", "wizardState.registrationData.primaryDAO.tokens.length")
+  get hasValidTokensDetailsCount() {
```
dkent600, line 23: should declare a return type

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -0,0 +1,78 @@
...
+      return this.form.validate()
+        .then(result =>
+          result.valid &&
+          this.hasValidTokensDetailsCount &&
+          Boolean(tokenDetailsValidationResults.filter(Boolean).length),
+        );
+    });
+  }
+
+  addToken() {
```
dkent600, line 60: declare a return type

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -0,0 +1,78 @@
...
+      address: "",
+      amount: undefined,
+      instantTransfer: 0,
+      vestedTransfer: 0,
+      vestedFor: undefined,
+      cliffOf: undefined,
+    });
+  }
+
+  deleteToken(token: IToken) {
```
dkent600, line 71: return type :-)

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
```diff
@@ -0,0 +1,122 @@
...
+      this.token.name = undefined;
+      this.token.logoURI = undefined;
+      this.token.symbol = undefined;
+      this.token.decimals = undefined;
+    }
+
+    this.infoSearched = true;
+  }
+
+  async save() {
```
dkent600, line 114: need a return type

src/services/TokenService.ts
dkent600: I asked to leave this.

src/wizards/tokenSwapDealWizard/components/stageButtons/stageButtons.html
```diff
@@ -9,8 +9,10 @@
     <pbutton
       if.bind="!showSubmit"
       type="primary"
-      click.delegate="wizardService.proceed(wizardManager)"
-      data-test="wizard-proceed-button">
+      click.delegate="proceed()"
+      data-test="wizard-proceed-button"
+      is-loading.bind="validating"
```
dkent600, line 14: why call this "is-loading" if is related to validating?  Why do we need this?

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
```diff
@@ -0,0 +1,158 @@
+<template>
```
dkent600, line 1: should not be located under "components" until we know it will be reused.

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
dkent600: I much prefer using div+grid.  I very rarely resort to using table.  Grid makes is more clear what behavior is required, and fewer permutations of technologies we have to work with, particularly in mobile responsiveness situations.

dkent600 (changes requested):
Needs a lot more testing.

Bad token url should give a validation message and not show the dead icon image.  This is Launch:

image

If the tokenInfo.name is not found, and I try typing in a name, this is what happens:

image

src/resources/elements/primeDesignSystem/pinput-group/pinput-group.scss
```diff
@@ -15,6 +16,12 @@ pinput-group {
     border-radius: $border-radius-normal;
   }
 
+  &.disabled {
+    .addon {
+      opacity: 0.43; // from the design system mock
```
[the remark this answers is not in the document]
IonelLupu: I am thinking this value is just a random value which will be changed to a proper design system value (when we will find it after the mockup is updated)

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.html
```diff
@@ -12,13 +12,10 @@
...
-             max.bind="max"
-             min.bind="min"
-             ref="rangeInput"
-             step.bind="step"
-             type="range"
-             value.bind="value"
+      <input
+        ref="rangeInput"
+        type="range"
+        value.bind="value"
```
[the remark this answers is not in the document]
IonelLupu: the min max values are by default from 0 to 100 and we don't have any settings to change that for the `prange-slide` component

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
```diff
@@ -10,33 +10,39 @@ export class PRangeSlider {
   @bindable leftLabel?: string;
```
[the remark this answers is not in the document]
IonelLupu: Where should I add this information? 

ps. I changed it because it didn't work as expected when it was created in the first place

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
IonelLupu: This is a bigger change since there are a few more pages that uses this. So yes, we need a different PR for it. (feel free to create an issue for it if you want )

src/resources/elements/primeDesignSystem/pform-input/pform-input.html
IonelLupu: There is an issue in Aurelia where the slot is not replaced by our custom HTML if the parent of the slot has an if statement. I couldn't make it work. In case of custom error messages, you will have to handle their visibility manually. We can have a call about it in case you know a solution to this problem. If so, then yes, the if.bind should be on the `p` tag.

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
IonelLupu: Any recommendation on where to put it? I couldn't find a place that shows the input value in the component but also changes the `this.left` property

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
IonelLupu: removed the div

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
IonelLupu: Yes. I wanted to do that but this validator uses the `this.tokenService` which I don't have access to in the `ValidationService` 😞

src/wizards/tokenSwapDealWizard/components/stageButtons/stageButtons.html
IonelLupu: I added this just to let the user know the app is doing something(validation in this case) before going to the next stage. If we don't have this loading icons, the user won't know why the "Proceed" button takes 2-3 seconds to change to the next stage. I am open to other suggestions.

hiaux0 (commented):
Peer reviewed together with @IonelLupu 
Code looks good overall 👍 

For me: test fnality

hiaux0 (commented):
Fnality
- the interaction between validation and switching-stages needs some refinement (future?)
  - addressed obvious flaw with red border around card

image

src/entities/DealRegistrationTokenSwap.ts
```diff
@@ -18,10 +18,18 @@ export enum Platforms {
...
-  balance: string,
   address: string,
+
+  name?: string,
+  symbol?: string,
+  decimals?: number,
+  logoURI?: string,
+
+  amount?: number;
+  instantTransfer: number
```
dkent600, line 29: amount, instantTransfer and vestedTransfer should be in wei, saved as a string (see, for example in launch, tokenDetails.maxSupply). Same for all eth/token amounts -- they must always be passed around in wei. numeric-input takes care of that, including converting to a string.

BTW I'd kinda prefer "instantTransferAmount" and "vestedTransferAmount"

src/resources/elements/primeDesignSystem/pform-input/pform-input.html
```diff
@@ -26,7 +26,7 @@
   <!-- Validation message section -->
   <p
     class="feedbackMessage ${ validationStateExists(validationState) ? 'feedbackMessageVisible': '' } ${validationState + 'Message'}">
-    ${validationMessage}
+    ${!inputReference.disabled ? validationMessage : ''}
```
dkent600, line 29: how about:

`${inputReference.disabled ? '' : validationMessage}`

src/resources/elements/primeDesignSystem/pinput-group/pinput-group.ts
```diff
@@ -1,9 +1,11 @@
-import { customElement, child } from "aurelia-framework";
+import { child, customElement } from "aurelia-framework";
 import "./pinput-group.scss";
+import { bindable } from "aurelia-typed-observable-plugin";
 
 @customElement("pinput-group")
 export class PInputGroup {
   element: Element;
+  @bindable.booleanAttr() disabled = false;
```
dkent600, line 8: doesn't need the "()"

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.html
```diff
@@ -10,8 +10,10 @@
     autocomplete.to-view="autocomplete"
   ></numeric-input>
 
-  <i if.bind="validationState === 'warning'" class="feedback-icon fas fa-exclamation-circle"></i>
-  <i if.bind="validationState === 'error'" class="feedback-icon fas fa-exclamation-triangle"></i>
-  <i if.bind="validationState === 'validating'" class="feedback-icon fas fa-circle-notch loading"></i>
+<template if.bind="!disabled">
```
dkent600, line 13: Would the validation stuff be better located in pform-input?

src/wizards/tokenSwapDealWizard/components/stageButtons/stageButtons.html
dkent600: I get it, ok

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
dkent600: ok

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.html
```diff
@@ -0,0 +1,184 @@
...
+          No valid image found at the provided logo URL
+        </span>
+        </div>
+      </div>
+      <pform-input data-test="proposal-token-amount-field"
+                   input-reference.bind="amountInput"
+                   label="Amount"
+                   label-description="You are able to edit the amount of tokens to be swapped before the execution phase."
+      >
+        <pinput-numeric
```
dkent600, line 55: note you can just say `not-wei`.  I'm not sure that `.bind="true"` is even valid since `true` isn't a view-model instance variable.  Should be `not-wei="true"`.  But just plain `not-wei` is nicest.

dkent600 (changes requested):
1) Type an invalid image url, then delete it.  This should say something about being required

image

2) Instant and vested should not be editable

image

3) Amount should be cleared if the token address changes

4) I just cleared (and also when reentered) the token address and this error message is still showing

image

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
dkent600: All you have to do is inject TokenService in ValidationService, no?

src/resources/elements/primeDesignSystem/pinput-numeric/pinput-numeric.html
IonelLupu: I tried that before and I tried it right now as well. It cannot be done because the icons have position: absolute and it won't look centered when you add more elements inside the pform-input "slot". Take this for example:
image

The icons will appear in the right hand side of the component 😢

dkent600 (changes requested):
1. entering an odd number like 5 results in 3 and 3 in the slider when set to 50%.  shouldn't it be 2.5 and 2.5?

2. typing numbers into either slider input doesn't have any effect on the slider input.  It should alter the slider position and the value of the other input.

3. entering an invalid value in a slider input should result in a validation error (like if the range is 10, enter 11).  

4. entering a cliff larger than the vesting period should display a validation error

5. I'd like to see more margin after the validation error messages.  Looks pretty cramped when a message is visible.

dkent600 (changes requested):
the slider left and right values need a tooltip that shows the value in full precision (like 2.12345), unless the spec calls for something different.

The cliff needs to be less than *or equal* to vesting period.

When I change the vesting period it should eliminate the cliff validation error

src/entities/DealRegistrationTokenSwap.ts
```diff
@@ -18,10 +18,18 @@ export enum Platforms {
 }
 
 export interface IToken {
-  name: string,
-  symbol: string,
-  balance: string,
   address: string,
+
+  name?: string,
```
dkent600, line 23: These shouldn't be optional?  Make sure they are initialized properly?

dkent600 (approved):

PR #186, opened 2022-02-18
chore(webpack): fix multiple emits

hiaux0:
## What was done
- no sourcemaps in dev mode (?)

## Testing
npm run start-dev
make a change in .html file

#### Before
grafik

#### After
should hot reload

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -92,7 +92,7 @@ module.exports = ({ production, extractCss, analyze, tests, hmr, port, host, } =
     path: outDir,
     publicPath: baseUrl,
     filename: production ? '[name].[chunkhash].bundle.js' : '[name].[fullhash].bundle.js',
-    sourceMapFilename: production ? '[name].[chunkhash].bundle.map' : '[name].[fullhash].bundle.map',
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -92,7 +92,7 @@ module.exports = ({ production, extractCss, analyze, tests, hmr, port, host, } =
     path: outDir,
     publicPath: baseUrl,
     filename: production ? '[name].[chunkhash].bundle.js' : '[name].[fullhash].bundle.js',
-    sourceMapFilename: production ? '[name].[chunkhash].bundle.map' : '[name].[fullhash].bundle.map',
+    sourceMapFilename: production ? '[name].[chunkhash].bundle.map' : undefined,
```
hiaux0, line 95: I compared with the aurelia-cli `webpack.config.js` file, and this was the difference (the cli one did not specify `sourceMapFilename` at all.

@dkent600 Do you have more insights (than me) about usage of `sourceMapFilename` in dev mode?

I tried to `throw new Error()` to see if we still get the filename in the trace.
Seems like we still do:
grafik

dkent600: no, sorry

dkent600 (approved):

PR #208, opened 2022-02-23
Update deal registration data to be initialized with undefined partnerDAO, clone existing deal registration data

paweljanicki:
## What was done
- DealRegistrationTokenSwap initialised with undefined PartnerDAO
- partnerDAO object with undefined values added to registration data for makeAnOffer and partneredDeal wizard types
- Clone deal registration data before modifying it


src/wizards/tokenSwapDealWizard/wizardManager.ts
```diff
@@ -161,7 +165,7 @@ export class WizardManager {
     }
 
     await deal.ensureInitialized();
-    return deal.registrationData;
+    return JSON.parse(JSON.stringify(deal.registrationData));
```
dkent600, line 168: I suspect that  `Object.assign({}, deal.registrationData)` would be more efficient.

src/wizards/tokenSwapDealWizard/wizardManager.ts
```diff
@@ -88,6 +88,10 @@ export class WizardManager {
     // if we are accessing an already existing deal, get its registration data
     this.registrationData = params.id ? await this.getDeal(params.id) : new DealRegistrationTokenSwap();
 
+    if (wizardType === WizardType.makeAnOffer || wizardType === WizardType.partneredDeal) {
```
dkent600, line 91: I don't understand why `wizardType === WizardType.partneredDeal`?  Can be documented?

In any case, needs to be on the clone, not the original.

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/wizardManager.ts
paweljanicki: It's because when creating new DealRegistrationTokenSwap() it has partnerDAO set to undefined. We need to change it to an actual `IDAO` object, otherwise partnerDao stage is broken

src/wizards/tokenSwapDealWizard/wizardManager.ts
paweljanicki: I've testing it and Object.assign doesn't deep clone. I've checked on MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#warning_for_deep_clone

dkent600 (changes requested):
Changes as discussed

PR #213, opened 2022-02-24
validate on previous and stepper change

dkent600:
## What was done

Validate the current wizard stage when the user navigates via the stepper or the previous button, but don't block navigation if the stage is invalid.

#### After


paweljanicki (approved):
Looks good

PR #241, opened 2022-03-03
fixed pselect behavior when used with and empty string instead of undefined

IonelLupu:
## What was done
fixed the issue where the pselect was marked with an error when used with an empty string value inside a pform-input


dkent600 (approved):

BartuCulha:
This feels like a wrong ticket to place this issue but issue seemed familiar. Let me know if you want me to create a separate ticket for this. @IonelLupu 


Steps to reproduce:
-Navigate to: https://prime-deals-staging.vercel.app/initiate/token-swap/partnered-deal/primary-dao
-Type anything to the representative address input field
-Delete what you have typed

Result:
The error message will stay there even if the string length value is equal to 0
Screenshot 2022-03-04 at 10-54-27 Create a Partnered Deal Prime Deals

Expected: 
Reset the error.

PR #279, opened 2022-03-07
prevented the user to proceed when the registration wizard had unsaved data in step 5 and 6

IonelLupu:
## What was done
Prevented the user to proceed when the registration wizard had unsaved data in step 5 and 6

### DIFF
--- a/src/wizards/services/WizardService.ts
+++ b/src/wizards/services/WizardService.ts
@@ -88,11 +88,11 @@ export class WizardService {
     this.router.navigate(this.getWizardState(wizardManager).cancelRoute);
   }
 
-  public async proceed(wizardManager: any): Promise<void> {
+  public proceed(wizardManager: any): Promise<void> {
     const wizardState = this.getWizardState(wizardManager);
     const indexOfActive = wizardState.indexOfActive;
 
-    await this.goToStage(wizardManager, indexOfActive + 1, true);
+    return this.goToStage(wizardManager, indexOfActive + 1, true);
   }
 
   public previous(wizardManager: any): void {
--- a/src/wizards/services/WizardService.ts
+++ b/src/wizards/services/WizardService.ts
@@ -92,7 +92,7 @@ export class WizardService {
     const wizardState = this.getWizardState(wizardManager);
     const indexOfActive = wizardState.indexOfActive;
 
-    this.goToStage(wizardManager, indexOfActive + 1, true);
+    await this.goToStage(wizardManager, indexOfActive + 1, true);
   }
 
   public previous(wizardManager: any): void {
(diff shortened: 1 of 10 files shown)


src/wizards/services/WizardService.ts
```diff
@@ -92,7 +92,7 @@ export class WizardService {
     const wizardState = this.getWizardState(wizardManager);
     const indexOfActive = wizardState.indexOfActive;
 
-    this.goToStage(wizardManager, indexOfActive + 1, true);
+    await this.goToStage(wizardManager, indexOfActive + 1, true);
```
dkent600, line 95: remove `await` and `async`.  They aren't needed and cause confusion.

src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
```diff
@@ -2,25 +2,31 @@ import { IBaseWizardStage, IStageMeta } from "../../dealWizardTypes";
...
   }
 
   activate(stageMeta: IStageMeta): void {
     this.wizardManager = stageMeta.wizardManager;
     this.wizardState = this.wizardService.getWizardState(this.wizardManager);
 
-    this.wizardService.registerStageValidateFunction(this.wizardManager, () => areFormsValid(this.termsForms));
+    this.wizardService.registerStageValidateFunction(this.wizardManager, async () => {
+      this.checkedForUnsavedChanges();
+      return await areFormsValid(this.termClauses.map(viewModel => viewModel.form)) && !this.hasUnsavedChanges;
```
dkent600, line 28: remove `await` and `async`

dkent600 (changes requested):
Bug:

1. Click Proceed
2. Type into clause
3. Click Save
4. Delete text in clause
5. Click Save

Result: nothing happens
Expect: Clause to be saved.

6. Create new clause
7. type text into it
8. Click Save

Result: nothing happens
Expect: Clause to be saved.

src/wizards/services/WizardService.ts
IonelLupu: Actually, they are because the proceed button uses it to show the loading icon. If we don't show the loading icon, the user won't know the Proceed button is doing something. The loading was added because Stage 5 takes about 2-3 sec to validate

src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
IonelLupu: it is needed in order to make the expression return a boolean with the expected values

dkent600: Sorry you are right, I wasn't accounting for the rest of the expression.

src/wizards/services/WizardService.ts
dkent600: I don't see how this is correct.  The caller is responsible for doing the await.

dkent600 (changes requested):

dkent600 (approved):

PR #283, opened 2022-03-07
fixed issue where tooltip was changing the element's style

IonelLupu:


src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
```diff
@@ -14,7 +14,7 @@ export class PTooltip {
   private tooltip: Instance;
 
   constructor(private element: Element) {
-    this.tooltip = tippy(this.element, {theme: "prime-design-system"});
+    this.tooltip = tippy(this.element, {theme: "prime-design-system", appendTo: document.body});
```
dkent600, line 17: See https://atomiks.github.io/tippyjs/v6/all-props/#appendto

Are you sure this works for all cases?  Should it be hard-coded as it is?  How about make it be a bindable string that is a selector, where the default is "body"?  But we need to make sure this works for all of the tooltip cases.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
dkent600: could also be bindable to all three legal datatypes

src/resources/elements/formattedNumber/formattedNumber.html
```diff
@@ -1,5 +1,6 @@
 <template>
-  <span ptooltip="content.bind: tooltip; placement.bind: placement; interactive: true; visible.bind: !hideTooltip">
+  <span
+    ptooltip="content.bind: tooltip; placement.bind: placement; interactive: true; visible.bind: !hideTooltip; append-to.bind: documentBody">
```
dkent600, line 3: I think it should work to simply be able to supply "body"

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
```diff
@@ -10,6 +10,7 @@ export class PTooltip {
   @bindable.boolean interactive = false;
   @bindable.boolean visible = true;
   @bindable.boolean allowHtml = false;
+  @bindable appendTo: Element;
```
dkent600, line 13: @bindable appendTo: Element | string;

src/resources/elements/formattedNumber/formattedNumber.ts
```diff
@@ -22,6 +22,8 @@ export class FormattedNumber {
   @bindable.booleanAttr public hideTooltip = false;
   @bindable.booleanAttr public thousandsSeparated = false;
 
+  documentBody = document.body;
```
dkent600, line 25: can delete this if we just supply "body"?

dkent600 (approved):

PR #298, opened 2022-03-09
Firestore service

paweljanicki:
#### After

### DIFF
--- a/src/services/FirestoreService.ts
+++ b/src/services/FirestoreService.ts
@@ -1,0 +1,28 @@
+import { addDoc, collection, setDoc, doc, query, where, getDocs } from "firebase/firestore";
+import { IDealRegistrationTokenSwap } from "entities/DealRegistrationTokenSwap";
+import { firebaseDatabase } from "./FirebaseService";
+
+const dealsCollection = "deals";
+
+export class FirestoreService {
+  public async createTokenSwapDeal(registrationData: IDealRegistrationTokenSwap) {
+    await addDoc(collection(firebaseDatabase, dealsCollection), {
+      registrationData,
+      discussions: ["qwe"],
+      votes: ["asd"],
+    });
+  }
+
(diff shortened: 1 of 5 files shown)


firebaseFunctions/src/types.ts
```diff
@@ -0,0 +1,112 @@
...
+interface IClause {
+  id: string,
+  text: string,
+}
+
+interface ITerms {
+  clauses: Array<IClause>,
+}
+
+export interface IDealRegistrationTokenSwap {
```
IonelLupu, line 66: Maybe there is a way to have a "Shared" directory where we can store interfaces used both by frontend and backend?

src/services/utils.ts
```diff
@@ -181,4 +181,9 @@ export class Utils {
 
     documentElement?.style.setProperty(varName, value);
   }
+
+  public static uniqById<T>(collection: Array<T & {id: string}>): Array<T & {id: string}> {
```
IonelLupu, line 185: I would add a type condition to this generic so that only objects that have an actual `id` property can be passed:
```
public static uniqById<T extends {id: string}>(collection: Array<T>): Array<T> {
```

src/services/FirestoreService.ts
```diff
@@ -0,0 +1,309 @@
...
+    );
+  }
+
+  /**
+   * Turns Firestore QuerySnapshot into a collection
+   *
+   * @param querySnapshot QuerySnapshot<DocumentData>
+   * @returns IFirebaseDocument[]
+   */
+  private getDocumentsFromQuerySnapshot(querySnapshot: QuerySnapshot<DocumentData>): Array<IFirebaseDocument> {
```
IonelLupu, line 291: maybe we can add this to the "Shared" directory as well (only if the "Shared" directory idea works in the first place 😄 )

IonelLupu (changes requested):

IonelLupu (approved):

dkent600 (changes requested):
Sending back to be completed

dkent600 (approved):

PR #304, opened 2022-03-09
Fix/submit stage etherscanlink

hiaux0:
⚠️ Note, that this PR goes into the submit-stage branch

## What was done
- add etherscanlink
grafik

- adjust width
before:
grafik
after:
grafik

- [x] tooltip width still needs fix 
BEFORE:
grafik

AFTER:
grafik

### DIFF
--- a/src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
+++ b/src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
@@ -3,6 +3,9 @@ import "./ptooltip.scss";
 import tippy, { Instance, Placement, Props } from "tippy.js";
 import { bindable } from "aurelia-typed-observable-plugin";
 
+/** Value found by getting width of 42 'M's of font size 14px. */
+const MAX_WIDTH = 560;
+
 @customAttribute("ptooltip")
 export class PTooltip {
   @bindable({primaryProperty: true}) content: string;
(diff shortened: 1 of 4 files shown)


src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
```diff
@@ -15,7 +15,12 @@ export class PTooltip {
   private tooltip: Instance;
 
   constructor(private element: Element) {
-    this.tooltip = tippy(this.element, {theme: "prime-design-system"});
+    /** Value found by getting width of 42 'M's of font size 14px. */
+    const maxWidth = 560;
```
IonelLupu, line 19: make it a bindable just in case someone has a really special use case where they want a bigger tooltip

IonelLupu (changes requested):

IonelLupu (approved):

src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
```diff
@@ -11,11 +14,15 @@ export class PTooltip {
   @bindable.boolean visible = true;
   @bindable.boolean allowHtml = false;
   @bindable appendTo: Props["appendTo"] | string;
+  @bindable.number maxWidth: number = MAX_WIDTH;
```
dkent600, line 17: why did you make this a bindable?  We have no current use for it.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/ptooltip/ptooltip.ts
hiaux0: ok will revert

cc @IonelLupu

IonelLupu: I was thinking this is needed in case we will have a tooltip that will exceed this size in the future. Just to be safe. @dkent600

dkent600: Generally best practice is not to code for possible future cases unless we are confident they will be needed.  @IonelLupu

dkent600 (approved):

PR #306, opened 2022-03-09
preserving terms stage clauses view mode

IonelLupu:


src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
```diff
@@ -53,4 +58,9 @@ export class TermsStage implements IBaseWizardStage {
   checkedForUnsavedChanges() {
     this.hasUnsavedChanges = this.termClauses.filter(viewModel => viewModel.viewMode === "edit").length > 0;
   }
+
+  private getDefaultTermsViewModes(wizardType: WizardType): EditingCard["viewMode"][] {
+    const isACreateWizard = [WizardType.createOpenProposal, WizardType.createPartneredDeal].includes(wizardType);
```
hiaux0, line 63: Future:
Wonder if checks like this can go into WizardService, and if not, what could that class/collection be?
(currently I understand "Service" as a convenience class as well)

```ts
public isACreateWizard(wizardType): boolean {}
```

src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
```diff
@@ -13,6 +14,7 @@ export class TermsStage implements IBaseWizardStage {
 
   termClauses: TermClause[] = [];
   hasUnsavedChanges = false;
+  stageMetadata: Partial<{ termsViewModes: EditingCard["viewMode"][] }> = {};
```
hiaux0, line 17:
```suggestion
  stageMetadata: { termsViewModes?: EditingCard["viewMode"][] } = {};
```

hiaux0: Played around a bit.
Following makes the `this.stageSettings` type a bit clearer imo

grafik
```ts
type StageSettings = WizardManager["additionalStageMetadata"][number];
type TermsStageSettings = { termsViewModes?: EditingCard["viewMode"][] } & StageSettings

@autoinject
export class TermsStage implements IBaseWizardStage {
  public wizardManager: any;
  public wizardState: IWizardState<IDealRegistrationTokenSwap>;

  termClauses: TermClause[] = [];
  hasUnsavedChanges = false;
  stageSettings: TermsStageSettings = {};

  constructor(
    public wizardService: WizardService,
  ) {
  }

  activate(stageMeta: IStageMeta<StageSettings>): void {
```

hiaux0 (approved):
- Code looks good
- fnality: fixed the original issue

dkent600 (approved):

src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
IonelLupu: mhm. Yes. Maybe we should add the `StageSettings` type in some place because it will be used many times.

But, there is a catch: `TermsStageSettings`  with `StageSettings` gives access to all the stage's setting and not only to what the users defines in the stage. This means they can access everything and they also need to define the types of those.

I would go with just `type TermsStageSettings = { termsViewModes?: EditingCard["viewMode"][] }` for now until we get to a point where we need to access more than the stage supposed to access

PR #308, opened 2022-03-09
styles(terms): increase specificity

hiaux0:
## What was done
quick fix for styles overwritten

#### Before
grafik

#### After
grafik


dkent600 (changes requested):
Please connect to an issue

dkent600 (approved):

PR #309, opened 2022-03-10
Fix Terms proceed validation (related to Submit Stage 7)

hiaux0:
## What was done
When on Terms stage (6) for the first time, save then proceed did not work (it should have)
--> Add separate validation for Terms Stage (aka. the "last stage")

Caused by: d071143840c718e527abdf78ea4fe31e0eda0a47

## Testing
1. Have all other stages valid
2. Be on Terms Stage
3. Put valid clause input
4. Save Clause
5. Proceed

#### Before
Could not proceed to Submit

#### After
Can proceed

### DIFF
--- a/src/wizards/services/WizardService.ts
+++ b/src/wizards/services/WizardService.ts
@@ -91,11 +91,12 @@ export class WizardService {
     this.router.navigate(this.getWizardState(wizardStateKey).cancelRoute);
   }
 
-  public proceed(wizardStateKey: WizardStateKey): Promise<void> {
+  public async proceed(wizardStateKey: WizardStateKey): Promise<void> {
     const wizardState = this.getWizardState(wizardStateKey);
     const indexOfActive = wizardState.indexOfActive;
 
-    if (!this.checkCanProceed(indexOfActive)) {
+    const canProceed = await this.checkCanProceed(wizardState, indexOfActive);
+    if (!canProceed) {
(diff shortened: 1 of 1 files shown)


src/wizards/services/WizardService.ts
```diff
@@ -186,12 +192,16 @@ export class WizardService {
...
 
-  private checkCanProceed(indexOfActive: number): boolean {
+  private async checkCanProceed(wizardState: any, indexOfActive: number): Promise<boolean> {
     const visibleStagesCount = this.getVisibleStages().length;
     const isLastStage = visibleStagesCount === indexOfActive + 1;
     /** Proceed just fine when not last stage. */
     if (!isLastStage) return true;
 
+    if (isLastStage) {
+      await this.validateActiveStage(wizardState, indexOfActive );
```
hiaux0, line 202: @dkent600 
Problem with this:
`validateActiveStage` will be called twice for the Terms Stage
Once: `checkCanProceed` (because proceeding from the "last stage" requires all other stages to be valid, cf. Launch)
Twice: `goToStage`

Imo, negligible.
(if not, I would have to refactor `goToStage` a bit, as I see it)

dkent600: I suspect we can live with that.

hiaux0:
done by d9251d81a216dc0d11aa3432aacf94df151f47ff

PR #320, opened 2022-03-11
fix(tokenDetails): only add default, when there is non

hiaux0:
## What was done
- add check to only add, when array is empty

## Testing
see ticket

#### Before
upon (re-) opening of token details stage, a new token was added

#### After
only add token, when array is empty

### DIFF
--- a/src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
+++ b/src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
@@ -128,8 +128,12 @@ export class TokenDetailsStage {
 
   private addDefaultValuesToRegistrationData(wizardType: WizardType) {
     if (wizardType === WizardType.createPartneredDeal) {
-      this.addToken(this.wizardState.registrationData.primaryDAO.tokens);
-      this.addToken(this.wizardState.registrationData.partnerDAO.tokens);
+      if (this.wizardState.registrationData.primaryDAO.tokens.length === 0) {
+        this.addToken(this.wizardState.registrationData.primaryDAO.tokens);
+      }
+      if (this.wizardState.registrationData.partnerDAO.tokens.length === 0) {
+        this.addToken(this.wizardState.registrationData.partnerDAO.tokens);
+      }
     }
   }
 }


IonelLupu (approved):

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -127,8 +127,12 @@ export class TokenDetailsStage {
   }
 
   private addDefaultValuesToRegistrationData(wizardType: WizardType) {
-    if (wizardType === WizardType.createPartneredDeal) {
+    if (wizardType !== WizardType.createPartneredDeal) return;
```
hiaux0, line 130: @dkent600 
For the "Make an offer", we should also add default tokens in the Token details stage (because we do so for the create a partnered deal too).

SOT: Make an offer 
> Just like Partnered Deal Registration Wizard except:

https://www.notion.so/primedao/Registration-Wizard-74901348d4a443a6891ad16fa1c0b42d#d1d7b1ebc7264bdc9e2b4d05541efa5e

Currently, we don't.
image

The change would be just to add `WizardType.makeAnOffer` to the if check as well

dkent600: > @dkent600
> 
> **Q: Do the following change here, or in another ticket?** (The change would be just to add `WizardType.makeAnOffer` to the if check as well)
> 
> For the "Make an offer", we should also add default tokens in the Token details stage (because we do so for the create a partnered deal too).
> 
> SOT: Make an offer
> 
> > Just like Partnered Deal Registration Wizard except:
> 
> https://www.notion.so/primedao/Registration-Wizard-74901348d4a443a6891ad16fa1c0b42d#d1d7b1ebc7264bdc9e2b4d05541efa5e
> 
> Currently, we don't. image

separate ticket please

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
hiaux0: @dkent600 

> separate ticket

> changes requested

To be clear, I did not add the "make an offer" part to this ticket.
Was the change request about that?

dkent600, line 130: why are you changing this logic?  The original line is better in being simpler and not adding an additional point of return from the function.

dkent600: got it

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
hiaux0: reverted

dkent600 (approved):

PR #328, opened 2022-03-11
Clauses and Tokens should not show Delete button when there is only one

hiaux0:
## Testing
have only 1
- clause
- token (in token details stage)

#### Before
- showed delete button

#### After
hides it

- clause
image

- tokenDetail
image

## Notes
Still shows button for 2+
image


src/resources/elements/editingCard/editingCard.ts
```diff
@@ -7,6 +7,7 @@ export class EditingCard {
   @bindable onEdit: () => boolean | Promise<boolean>;
   @bindable onSave: () => boolean | Promise<boolean>;
   @bindable({defaultBindingMode: bindingMode.twoWay}) viewMode: "edit" | "view" = "edit";
+  @bindable.boolean hideDeleteButton: boolean;
```
dkent600, line 10: Can we explicitly set the default to false?

dkent600 (changes requested):

dkent600 (approved):

hiaux0:
Conflicts resovled

Issue #318, opened 2022-03-11
'Time Passed' should show `<1min` instead of seconds and update constantly.


giladt:
## Current
'Time since' appears in `Discussions List` ('last activity') and in `Comment Thread` ('created').
<img src="https://images.zenhubusercontent.com/60c89178ac3fd01fa6c9ba4a/95e025b3-b19f-4b27-8f16-402b7b92651c" width="200"> <img src="https://images.zenhubusercontent.com/60c89178ac3fd01fa6c9ba4a/44ae01be-f26f-429f-bf53-102881373a5d" width="200">

## Expected:
- [ ] It should show `<1min` instead of counting seconds.

<img src="https://images.zenhubusercontent.com/60c89178ac3fd01fa6c9ba4a/e47db6c0-c584-4681-b701-28df689b8e05" width="200">

- [ ] It should update counting as time goes by.

<img src="https://images.zenhubusercontent.com/60c89178ac3fd01fa6c9ba4a/0a46d555-eb10-4c58-8f07-b21e578ad470" width="200">

curvelabs-dev:
Just FYI Chad is changing these to "1 min" (with a space).  Is also changing "1m" to "1 mo" (to distinguish between minute and month.  Will also be making them nowrap.

PR #330, opened 2022-03-11
#318 - 'Time Passed' should show `<1min` instead of seconds and update constantly.

giladt:
## What was done
`Last activity` shows '<1min' if less than a minute has passed since the last activity has happened.
Applies to `discussions list` activity column and `single comment` time since created.

Preview:
https://prime-deals-dapp-git-fix-discussions-list-shou-3dd68b-curvelabs.vercel.app/deal/open_deals_stream_hash_2


src/dealDashboard/discussionsService.ts
```diff
@@ -160,6 +170,7 @@ export class DiscussionsService {
         createdBy,
         createdAt: new Date(),
         modifiedAt: new Date(),
+        lastModified: "<1min",
```
dkent600, line 173: Can please be "< 1 min"?

src/services/DateService.ts
```diff
@@ -45,8 +45,8 @@ export class DateService {
       relativeTime: {
         future: "in %s",
         past:   "%s ago",
-        s:  "%dsec",
-        ss: "%dsec",
+        s:  "<1min",
```
dkent600, line 48: Please add spaces to all of these between the number and the unit.  "m" should be "mo"

dkent600 (changes requested):

src/services/DateService.ts
giladt: done

src/dealDashboard/discussionsService.ts
giladt: done

dkent600 (approved):

Issue #332, opened 2022-03-11
s


BartuCulha:

hiaux0:
@BartuCulha 
1. address should change color to white
2. ? Add tooltip to address

Can you clarify, if you expect the 2. point as well?

curvelabs-dev:
I actually already worked this

PR #335, opened 2022-03-11
#332 Use etherscanlink in wallet menu

hiaux0:
## What was done
- [ ] awaiting product input about what is expected

#### After
image


hiaux0:
@dkent600 
Should we remove the right copy button?

image

dkent600 (changes requested):
We should not be using etherscanlink here.  We should simply be doing what Bartu asked for, which asks to change the button color on hover over the button (and change the title of this PR :-) ).

PR #351, opened 2022-03-11
Made the home page mobile responsive

IonelLupu:
## What was done
Mock: https://www.figma.com/file/hTLvqCEzki3WoG5uJbtynw/Prime-Deals?node-id=7557%3A116288
image

#### After

### DIFF
--- a/src/app.scss
+++ b/src/app.scss
@@ -3,9 +3,9 @@
 .appWrapper {
   display: grid;
   grid-template-columns:
-    minmax(#{$MinContentBodyPaddingLeftRight}, 1fr)
-    #{$ContentBodyWidth}
-    minmax(#{$MinContentBodyPaddingLeftRight}, 1fr);
+    minmax($ContentMinLeftRightPadding, auto)
+    minmax(auto, $MaxContentBodyWidth)
+    minmax($ContentMinLeftRightPadding, auto);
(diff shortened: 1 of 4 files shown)


hiaux0:
**! Note !
Code looks straightforward.
Fnality: I started to compare with mocks (see below), but realized, we don't have to go pixel perfect, so all of that below is just optional.**

---
---

### Wallet menu right margin
image

### Navbar margin
image

### View all Deals left margin
image

### View all Deals margin bottom
image

IonelLupu:
@hiaux0  Updated the design. Please take a look again. 

These is still one weird issues about the left-right spacing of the entire app (your "View all Deals left margin" image) which shows pretty much correctly on my side:
image


It seems that on your browser the space is way smaller. I will have to inspect that

hiaux0:
> It seems that on your browser the space is way smaller. I will have to inspect that

Don't worry I'd say. Somehow my browser has weird glitches all around the place.

dkent600 (changes requested):
1) I don't think these cards should be single-column at this breakpoint around 1000px:

image

2) Tabs should have greater margins at this point (more centered):

image

dkent600 (changes requested):
Changes per our discussion today

dkent600 (approved):

Issue #357, opened 2022-03-14
FormattedNumber's tooltip should also use "," for long numbers


hiaux0:
Relates to #344
in Token Details and Submit stage for the token information card.

Idea: Any long number should handle this by default (in `<formatted-number>`)

![image](https://user-images.githubusercontent.com/30693990/158166603-2359914b-38b7-4269-9036-43e4ccfb6111.png)

curvelabs-dev:
I would prefer to not refactor code to make it be the default.  You can supply `| withCommas`.

hiaux0:
`Optional` label reason

Show the "raw" number for user to copy paste

curvelabs-dev:
I don't support making this be the default.  Should probably close this ticket.

PR #359, opened 2022-03-14
#357+358 - tooltip comma

hiaux0:
#### After
- 358
image

### DIFF
--- a/src/wizards/tokenSwapDealWizard/stages/submitStage/daoInformation/daoInformation.html
+++ b/src/wizards/tokenSwapDealWizard/stages/submitStage/daoInformation/daoInformation.html
@@ -110,7 +110,8 @@
             </p>
           </div>
 
-          <div class="detailsCell">
+          <let token-vested-for-tooltip.bind="token.vestedFor | withCommas"></let>
+          <div class="detailsCell" ptooltip.bind="tokenVestedForTooltip + ' seconds'">
             <p>Vested for</p>
             <p>
               <template if.bind="token.vestedFor"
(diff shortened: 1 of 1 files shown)


src/resources/elements/formattedNumber/formattedNumber.ts
```diff
@@ -54,7 +54,9 @@ export class FormattedNumber {
   }
 
   @computedFrom("_value")
-  private get tooltip():string {
-    return this._value?.toString(10);
+  private get tooltip(): string {
+    const withCommasValueConverter = this.container.get(WithCommasValueConverter);
+    const formatted = withCommasValueConverter.toView(this._value?.toString(10));
+    return formatted;
```
hiaux0, line 60: https://github.com/PrimeDAO/prime-deals-dapp/issues/357#issuecomment-1067391004
> I would prefer to not refactor code to make it be the default. You can supply | withCommas to formattedNumber.

In that case, can you help me with the below code?
I can't get that to work with the `ethwei` value converter

```html
<formatted-number
  if.bind="token.instantTransferAmount"
  thousands-separated
  value.to-view="token.instantTransferAmount | ethwei | withCommas"      <---- here
></formatted-number>
```

hiaux0: A bit of debugging shows, that there may be a big in `FormattedNumber.ts`
Somewhere in this if-check
https://github.com/PrimeDAO/prime-deals-dapp/blob/52ec5ee001e77e9f00260b4dc94783f8ba637237/src/resources/elements/formattedNumber/formattedNumber.ts#L41

image

dkent600 (changes requested):

dkent600 (approved):

PR #368, opened 2022-03-14
Build out all deals page

vipergtsrz:
Built out the all deals page
Currency formatter was added with a optional decimal places param.

Unit tests were created for the all deals page

### DIFF
--- a/src/deals/list/deals.ts
+++ b/src/deals/list/deals.ts
@@ -141,14 +141,6 @@ export class Deals {
   }
 
   /**
-   * This event is subscribed on the connect button and will open the meta mask prompt
-   * @fires {@link EventType.AccountConnect}
-   */
-  private onConnect(): void {
-    this.eventAggregator.publish("account.connect");
-  }
-
-  /**
    * Flips the toggle for whether or not deals for user are shown and also checks visbility
    */
   private toggleMyDeals(): void {
--- a/src/deals/list/deals.ts
+++ b/src/deals/list/deals.ts
@@ -115,7 +115,7 @@ export class Deals {
    * @param showMine Whether or not to show your current deals
    * @returns
    */
-  private getDealsForCardIndex(cardIndex: number, showMine: boolean) {
+  private getDealsForCardIndex(cardIndex: number, showMine: boolean) : DealTokenSwap[] {
     if (cardIndex === 0) {
       //open proposals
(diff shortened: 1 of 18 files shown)


src/resources/elements/ConnectButton/ConnectButton.ts
```diff
@@ -58,7 +60,9 @@ export class ConnectButton {
       this.txPhase = Phase.None;
       this.txReceipt = null;
     }));
-
+    this.subscriptions.push(this.eventAggregator.subscribe("account.connect", async () => {
```
dkent600, line 63: why is this needed?

dkent600 (changes requested):

src/deals/list/deals.ts
```diff
@@ -1,78 +1,161 @@
...
+    }
+  }
+
+  /**
+   * Returns a filtered set of deals based on criteria being passed in from the deal service
+   * @param cardIndex The current Tab
+   * @param showMine Whether or not to show your current deals
+   * @returns
+   */
+  private getDealsForCardIndex(cardIndex: number, showMine: boolean) {
```
dkent600, line 118: This should be implemented in the DealsService, not here.

src/resources/elements/ConnectButton/ConnectButton.ts
vipergtsrz: When the user's wallet isn't connected, the "onConnect" method in the deals.ts code calls 

`this.eventAggregator.publish("account.connect");`

Which triggers the exact same code as the 'onConnect' method inside the ConnectButton.ts. The code you're referring to is needed to allow the deals.ts code to connect their wallet without having to duplicate code.

src/deals/list/deals.ts
```diff
@@ -1,78 +1,161 @@
+import "./deals.scss";
+
+import { autoinject, computedFrom, singleton } from "aurelia-framework";
+
+import { DateService } from "services/DateService";
+import { DealService } from "services/DealService";
```
dkent600, line 6: This will create a module reference loop.

src/deals/list/deals.ts
dkent600: Also should have a return value

dkent600, line 118: All methods should declare their return type.

dkent600 (changes requested):
Sort by DAO and Title

dkent600:
When we change accounts while "Show My Deals Only", we should recompute the array

dkent600 (approved):

PR #389, opened 2022-03-16
383 discussion thread always shows is private when not connected

giladt:
## What was done
-[x] Fixed a bug when empty thread shows as private if no wallet is connected.
-[x] Fixed small typo, causing a deal is private discussion section show even if not authorized.
-[x] Converted the deal `open_deals_stream_hash_1` to private, for easier testing of private-deals functionality.

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -97,6 +97,16 @@ export class DiscussionThread {
     }
   }
 
+  @computedFrom("isLoading.discussions", "isMember", "threadComments.length")
+  private get noCommentsText(): string {
+    if (!this.isLoading.discussions && !this.threadComments.length) {
+      return (!this.isMember && this.deal.registrationData.isPrivate)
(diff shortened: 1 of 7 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -340,6 +350,6 @@ export class DiscussionThread {
   }
 
   private navigateTo() {
-    this.discussionId = null;
+    this.discussionId = "";
```
IonelLupu, line 353: I assume the empty string denotes the discussion is not selected. I would recommend we use `undefined` for when a discussion is not selected. An empty string might be confusing. (you can also remove the default empty string from the class property)

_Optional_

IonelLupu (approved):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
giladt: done

IonelLupu (approved):

dkent600 (approved):

Issue #386, opened 2022-03-16
Implement dealDashboard: Description section


curvelabs-dev:

hiaux0:
@dkent600 

I don't see anything else needed for this.
Can you point me to an open ticket?

Else, should be code complete

PR #396, opened 2022-03-16
#386 - Deal information

hiaux0:
## What was done
update dashboard skeleton with 
- deal privacy
- deal title 
- deal status (only text, no styling)
- read more

#### Not in this PR
- deal status (styling)
- deal activity (time-left)

## Testing
- choose a deal from the deal carousel from the home page
- Responsivness: into mobile view -> deal description longer than 250? -> read more button

#### After
grafik

### DIFF
--- a/src/dealDashboard/dealDescription/dealDescription.ts
+++ b/src/dealDashboard/dealDescription/dealDescription.ts
@@ -16,6 +16,8 @@ export class DealDescription {
   private descriptionText: string | undefined;
   private originalDescriptionText = "";
 
+  private resizeWatcher;
+
   @computedFrom("descriptionText")
   get showReadMoreButton() {
     return this.descriptionText.length < this.originalDescriptionText.length;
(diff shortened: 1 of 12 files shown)


src/dealDashboard/dealDescription/dealDescription.ts
```diff
@@ -1,9 +1,42 @@
...
+
+  @computedFrom("descriptionText")
+  get showReadMoreButton() {
+    return this.descriptionText.length < this.originalDescriptionText.length;
+  }
+
+  bind() {
+    this.originalDescriptionText = this.deal.registrationData.proposal.description;
+
+    if (MobileService.isMobile()) {
```
hiaux0, line 27: Currently, when one resizes the browser (eg. from 375px to +376px), then nothing will change.

@dkent600 Should I also add a viewport listener?

dkent600 (approved):
Interesting addition of MobileService

Issue #414, opened 2022-03-17
[TokenDetails] Wrong address shows token information sub form


hiaux0:
1. Token Details stage
2. copy wallet address into Address field
3. correctly show validation error
4. switch to another stage, and come back
5. Bug: see screenshot

![image](https://user-images.githubusercontent.com/30693990/158904947-d9002f30-99cb-4a27-ad5f-489a28236bed.png)

BartuCulha:
there's an another similar issue:

0x43D4A3cd90ddD2F8f4f693170C9c8098163502ad

Token Details stage
copy this address: 0x43D4A3cd90ddD2F8f4f693170C9c8098163502ad
Change the last letter from "d" to anything else
switch to another stage, and come back
Bug: see screenshot

![Create-a-Partnered-Deal-Prime-Deals(1)](https://user-images.githubusercontent.com/24368665/159707768-cab521bb-eb36-4558-8192-aa3d33f51905.png)

curvelabs-dev:
@hiaux0 I would add that it is not consistent with other pages that we would be seeing the red stuff in this specific case of returning to an invalid page from another page.

hiaux0:
@BartuCulha 

Not reproducible on staging
https://prime-deals-staging.vercel.app/initiate/token-swap/open-proposal/token-details
(firefox and chromium)

![Peek 2022-03-23 23-29](https://user-images.githubusercontent.com/30693990/159806830-2738be77-a7b5-414c-9bca-8bf00a0b3b39.gif)

BartuCulha:
I believe you have reproduced the issue. Mind the adress is invalid and you are able to see d2d token information

PR #418, opened 2022-03-18
#414 - Token details wrong address still shows token information sub form

hiaux0:
## What was done
remove default decimals
why?
- causes the bug (this was the quickest fix, else have to rethink logic when to show token details metadata)
- 413 says "The default of 18 is deprecated and will soon be replaced by an exception", so will be deprecated anyway? (that's why I chose this approach)
   - #421 removes the default decimals

#### After
should not show token metadata

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -68,6 +68,10 @@ cy.get("[data-test='open-proposal-button']").click()
 cy.url().should("include", "deals/open");
 ```
 
+### Tags
+- `@focus` - only run focused Scenario in .feature file
+- `@regression` - Indicate Scenario is to cover a regression in the code
+
 ### Tooling
 - [VSCode Cucumber Autocomplete Extension](https://github.com/alexkrechik/VSCucumberAutoComplete#settings-example)
 
--- a/cypress/integration/common/deal-wizard.ts
+++ b/cypress/integration/common/deal-wizard.ts
@@ -19,6 +19,13 @@ const stageTitlesToURLs = {
 export let sectionTitle = "";
 
 export function withinWizardSection() {
+  if (sectionTitle === "") {
+    const sectionErrorMessage = "Please specify the section you are targeting in the Wizard.\n"
+      + "Use the following step:\n\n"
+      + "  Given I want to fill in information for the \"<sectionName>\" section";
+    throw Error(sectionErrorMessage);
+  }
+
   return cy.contains("[data-test='section-title']", sectionTitle).parents(".stageSection");
 }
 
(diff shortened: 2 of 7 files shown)


src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -101,7 +101,7 @@ export class TokenDetailsStage {
 
       name: "",
       symbol: "",
-      decimals: 18,
+      decimals: NaN,
```
dkent600, line 104: why not undefined instead of NaN?

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
hiaux0: type defintion of `decimals` is only `number`.
I'll change the type to `number | undefined`.

dkent600: I wonder if it would be best to just work with the TokenService system that sets 0 as the default, consider 0 to be not set, and don't allow it to go undefined when the user clears the field?

dkent600 (changes requested):
re-submitting review to send this back

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
hiaux0: One problem with that: Cannot distinguish decimal value to disable Amount field

Background: Want `decimals = undefined` to be able to disable the Total amount field
_Relevant quote from Telegram:_
> Does this mean that, on the token details card, we have to have the "Amount" disabled while waiting for the 'Decimals" to load from the internet after the user fills in the "Address"?

> yes, which seems reasonable anyway

With `decimals` defaulting the 0, we can have a case, where
1. User fills in "Total amount" value --> decimal is 0
2. User fills in "Token address" --> decimal changes to eg. 18
3. --> "Total amount" value changes, because `decimal` changed (if my knowledge is correct)

dkent600 (changes requested):
Sending back per our conversations

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -101,7 +102,7 @@ export class TokenDetailsStage {
 
       name: "",
       symbol: "",
-      decimals: 18,
+      decimals: TokenService.DefaultDecimals,
```
hiaux0, line 105: use default value

hiaux0: There are 2 PRs, that do the same change.
Reason: both PRs needed to change this to fix underlying issue

image

dkent600 (approved):

PR #448, opened 2022-03-24
422 - Privacy label depending on wizard type

hiaux0:
## What was done
- e2e (for now disabled, because don't have full registration wizard flow yet
- change text based on wizard type

#### After
Open proposal

Partnered 

and Make offer


src/wizards/tokenSwapDealWizard/stages/submitStage/submitStage.ts
```diff
@@ -46,4 +48,23 @@ export class SubmitStage {
       this.eventAggregator.publish("handleFailure", `There was an error while creating the Deal: ${error}`);
     }
   }
+
+  private setPrivacyText(wizardType: WizardType): void {
```
dkent600, line 52: Is better to keep the text in the HTML

dkent600 (changes requested):

dkent600 (approved):

PR #470, opened 2022-03-25
Update discussions with updated data source deals

paweljanicki:
## What was done
- Replace old IDataSourceDeals in Discussions
- Remove ceramicMock
- Change how discussions service interacts with data source

#### After


src/dealDashboard/discussionsService.ts
```diff
@@ -232,7 +230,12 @@ export class DiscussionsService {
     if (timestamp)
       this.discussions[discussionId].modifiedAt = timestamp;
 
-    this.dataSourceDeals.update(discussionId, JSON.stringify(this.discussions[discussionId]));
+    this.dataSourceDeals.addClauseDiscussion(
```
dkent600, line 233: This is supported in DealTokenSwap.  Would be better to use that.  Note that DealTokenSwap is maintaining clauseDiscussions which I assume we need to keep in sync because the discussions components are using it.

src/entities/IDealTypes.ts
```diff
@@ -22,7 +22,7 @@ export interface IDealVotingSummary {
 export interface IDealTokenSwapDocument {
   id: string;
   registrationData: IDealRegistrationTokenSwap;
-  discussions: any; // TODO
+  clauseDiscussions: Record<string, any>; // TODO should be IDealDiscussion, but breaks firebaseFunctions build
```
dkent600, line 25: Why does it break?  Can we fix that by moving the IDealDiscussion interface to where are located all the others that define the document?

src/entities/IDealTypes.ts
```diff
@@ -45,11 +45,11 @@ export enum DealStatus {
 export interface IDeal {
   id: string;
   corrupt: boolean;
-  clauseDiscussions: Map<string, string>;
+  clauseDiscussions: Map<string, any>;
```
dkent600, line 48: use IDealDiscussion

src/entities/IDealTypes.ts
```diff
@@ -45,11 +45,11 @@ export enum DealStatus {
...
   registrationData: any;
   initialize(): Promise<void>;
   create<TDealDocumentType extends IDealTokenSwapDocument>(doc: TDealDocumentType): IDeal;
   ensureInitialized(): Promise<void>;
-  addClauseDiscussion(clauseId: string, discussionKey: string): Promise<void>;
+  addClauseDiscussion(clauseId: string, discussion: any): Promise<void>;
```
dkent600, line 53: use unknown instead of any?

dkent600 (changes requested):
Removing ceramicMock should not be part of this PR.

PR #471, opened 2022-03-25
469 - Fix DateService time ticks

hiaux0:
## What was done
Fix `largest2`

#### Before
"1 day, 0 seconds"

#### After
"1 day, 0 hours"

### DIFF
--- a/test/unit/services/DateService.spec.ts
+++ b/test/unit/services/DateService.spec.ts
@@ -1,0 +1,148 @@
+import { DateService, TimespanResolution } from "../../../src/services/DateService";
+
+type TimeArray = [
+  days: number,
+  hours: number,
+  minutes: number,
+  seconds: number
+];
+type TestData = [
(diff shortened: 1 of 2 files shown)


test/unit/services/DateService.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe("DateService", () => {
+  let dateService: DateService;
+  beforeAll(() => {
+    dateService = new DateService();
+  });
+
+  describe("#ticksToTimeSpanString", () => {
+    /* prettier-ignore */
+    const testData: TestData[] = [
```
hiaux0, line 23: @dkent600 @IonelLupu 

Could you also look over these test cases?
I went through them, and should be what we want now.

IonelLupu: yes, looks good

dkent600 (approved):

PR #476, opened 2022-03-28
Change votes structure inside votingSummary to a map

paweljanicki:


src/entities/IDealTypes.ts
```diff
@@ -1,15 +1,12 @@
 import { IDealRegistrationTokenSwap } from "./DealRegistrationTokenSwap";
 
-export interface IVoteInfo {
-  address: string;
-  vote: boolean;
-}
+export type IVotesInfo = Record<string, boolean>;
```
IonelLupu, line 3: votes can also be `null` when a representative didn't yet vote, right?

IonelLupu (changes requested):

src/entities/IDealTypes.ts
paweljanicki: Yes, should be update the type?

IonelLupu: yes please. It will help us when building some logic based on those

dkent600: I am not clear what is the issue here.  Has it been addressed?  @IonelLupu

IonelLupu: @dkent600  since @paweljanicki  is off, I fixed the type.

The issues was the type of the vote which is `boolean | null` in reality and not just `boolean`. I pushed a commit with the fix

IonelLupu (approved):

dkent600 (approved):

Issue #496, opened 2022-03-29
Grid columns should have column padding and display ellipses instead of clipping


curvelabs-dev:
Prime Launch provides ellipses and a bit of padding to make this row in All Deals look better:

![image.png](https://images.zenhubusercontent.com/61ae60c2002a061be6dc8669/10610e47-ee43-427d-b7b2-eccc96610582)

PR #498, opened 2022-03-29
Grid columns should have column padding and display ellipses instead of clipping

vipergtsrz:
Fix bug #496


vipergtsrz:
I also added logic to show a tooltip on hover if the content is cut off with the ellipsis

dkent600 (approved):

PR #500, opened 2022-03-29
E2e/pass firestore instance to tests

hiaux0:
#### After

### DIFF
--- a/cypress/integration/common/deal-api.ts
+++ b/cypress/integration/common/deal-api.ts
@@ -1,11 +1,76 @@
-import { Given } from "@badeball/cypress-cucumber-preprocessor/methods";
+import { IDealTokenSwapDocument } from "../../../src/entities/IDealTypes";
+import { E2eWallet } from "../tests/wallet.e2e";
+import { E2eNavigation } from "./navigate";
 
-export class E2eDealApi {
-  // public static getDeal(dealId) {
-  // }
+export class E2eDealsApi {
+  private static getDealService() {
+    // @ts-ignore - Hack to access firestore inside Cypress
+    return Cypress.firestoreDealsService;
+  }
 
+  public static getDeals(
+    address: string,
+  ): Cypress.Chainable<IDealTokenSwapDocument[]> {
+    cy.log("[Test] Navigate to home page, and wait for app boostrapping");
+    /**
+     * Need to have the app bootstrapped, in order to use firestore
+     */
+    E2eNavigation.navigateToHomePage();
+
+    return cy.then(async () => {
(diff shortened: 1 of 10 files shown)


cypress/integration/common/deal-api.ts
```diff
@@ -1,11 +1,76 @@
...
+import { E2eNavigation } from "./navigate";
 
-export class E2eDealApi {
-  // public static getDeal(dealId) {
-  // }
+export class E2eDealsApi {
+  private static getDealService() {
+    // @ts-ignore - Hack to access firestore inside Cypress
+    return Cypress.firestoreDealsService;
+  }
```
hiaux0, line 9: This is where `firestoreDealsService` is used

src/main.ts
```diff
@@ -61,7 +61,16 @@ export function configure(aurelia: Aurelia): void {
...
+        /**
+         * Mock wallet connection
+         */
         aurelia.use.singleton(EthereumService, EthereumServiceTesting);
+        /**
+         * Tests can directly access FirestoreDealsService.
+         * We want that to, eg. get dealIds from the dealsArray
+         */
+        const firestoreDealsService = aurelia.container.get(FirestoreDealsService);
+        (window as any).Cypress.firestoreDealsService = firestoreDealsService;
```
hiaux0, line 73: Where we pass `firestoreDealsService` into cypress

hiaux0 (commented):
And other fixes for the (previously) hard coded deal ids

paweljanicki:
Looks like it loads deals from Firestore correctly, and tests for "make an offer" wizard are opened with a URL containing existing dealId.

I suggest to use separate seed data for e2e tests. That way we avoid issues when someone updates seed data and it breaks the tests.

You could add following npm scripts, for updating seed data specific to e2e tests and starting emulators using it:
```
    "firebase-update-seed-data:e2e": "firebase emulators:export --force firebase_seed_e2e",
    "firebase-emulators:start:e2e": "firebase emulators:start --project=default --import=firebase_seed_e2e",
```

src/services/FirebaseService.ts
```diff
@@ -9,6 +8,8 @@ import { getFunctions, connectFunctionsEmulator } from "firebase/functions";
 import { Utils } from "services/utils";
 import { EventAggregator } from "aurelia-event-aggregator";
 
+type Address = string;
```
hiaux0, line 11: I would like to have `Address` be defined somewhere else than `EthereumService`.
Reason: I cannot import `FirebaseService` easily into Cypress, because it pulls in `EthService`, so quickest solution was to remove the dep on `EthService`.

In general, reducing the import footprint of each file might be a good idea overall

PR #509, opened 2022-03-30
E2e - Pass firestore instance to tests

hiaux0:
#### Services
- feat(e2e): pass FirestoreSerice to cypress - bd02fa4
- refac(e2e): Remove dep on EthereumService. Reason: - e01b751
- refac(services): Only show disclaimer when "Connect to a Wallet" is c… - fbbdc50

#### Cypress
- e2e(api): use FirestoreService in Cypress - b5c643c

#### Docs
- docs(cypress): page objects - d506d17 

## ! Note
There will be a follow PR. #511 
Reason: Split up PRs to show the diff better

## after
30.03.2022 - 8pm UTC+2 (CEST)
image

### DIFF
--- a/src/services/EthereumServiceTesting.ts
+++ b/src/services/EthereumServiceTesting.ts
@@ -1,6 +1,4 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
-const ADDRESS = "0xE834627cDE2dC8F55Fe4a26741D3e91527A8a498";
-
 /* eslint-disable no-console */
 import { ethers, Signer } from "ethers";
 import { BaseProvider, Web3Provider } from "@ethersproject/providers";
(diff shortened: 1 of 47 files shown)


src/services/FirebaseService.ts
```diff
@@ -1,4 +1,3 @@
-import { Address } from "services/EthereumService";
```
hiaux0, line 1: - Reason1: I cannot import FirebaseService easily into Cypress, because it pulls in EthService, so quickest solution was to remove the dep on EthService.
- Reason2: Easiest way, to not have too much trouble with webpack
        - In general, reducing the import footprint of each file might be a good idea overall

src/services/EthereumServiceTesting.ts
```diff
@@ -51,14 +49,10 @@ export class EthereumServiceTesting {
   ]);
 
   private async fireAccountsChangedHandler(account: Address) {
-    if (account && !(await this.disclaimerService.ensurePrimeDisclaimed(account))) {
```
hiaux0, line 54: - Not super happy about this change
- Reason: On first app load, we have this `EthereumService#connectToConnectedProvider` method, that checks for disclaimer. That hindered testing.
- Change seems minor, and rejecting Disclaimer will still block user --> TODO later: Add disclaimer test

hiaux0 (commented):
.

dkent600 (approved):

PR #526, opened 2022-04-01
Showing a popup notification after voting

IonelLupu:


src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -74,4 +76,9 @@ export class DealVotes {
   get statusText() {
     return this.everyTextCopy.find(textCopy => textCopy.condition());
   }
+
+  vote(value: boolean) {
+    this.eventAggregator.publish("showMessage", "Vote submitted");
```
dkent600, line 81: How it ought to work:

1) upon submission the vote button should show a spinny icon
2) put the vote call in try/catch
3) await deal.vote()
4) stop the spinning
4) if it hasn't crash then issue a nice popup notif "Your vote has been successfully submitted"
5) if it has crashed then a notif "Sorry, an error occurred submitting your vote"

src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -74,4 +76,9 @@ export class DealVotes {
...
+    return this.deal.vote(value);
```
dkent600, line 82: no point in returning anything here, so is best not to

dkent600 (changes requested):

src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -74,4 +79,24 @@ export class DealVotes {
...
+    this.declining = true;
+    await this.vote(true).finally(() => this.declining = false);
+  }
+
+  private async vote(value: boolean) {
+    try {
+      await this.deal.vote(value);
+      this.eventAggregator.publish("showMessage", "Your vote has been successfully submitted");
+    } catch (error) {
+      console.log("Voting error", error);
```
dkent600, line 98: This will not lint.  Use ConsoleLogService

src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
```diff
@@ -30,6 +30,7 @@
   user-select: none;
   color: $Neutral04;
   white-space: nowrap;
+  gap: 12px;
```
dkent600, line 33: we're not using this feature in this PR.  Do we need these changes?

src/dealDashboard/dealVotes/dealVotes.html
```diff
@@ -7,24 +7,30 @@ <h1 class="title heading heading3 gradient">Vote</h1>
...
           type="formfield"
-          click.delegate="deal.vote(false)"
+          click.delegate="declineDeal()"
         >
-          <i class="fas fa-times-circle"></i>
-          ${deal.representativeVote(ethereumService.defaultAccountAddress) === false ? 'Declined': 'Decline' }
+          <span class="icon is-loading" if.bind="declining">
+            <i class="spinner fas fa-sync-alt" data-test="loadingIcon"></i>
+          </span>
+          <i class="fas fa-times-circle" if.bind="!declining"></i>
```
dkent600, line 32: you can just use `else` here

src/dealDashboard/dealVotes/dealVotes.html
```diff
@@ -7,24 +7,30 @@ <h1 class="title heading heading3 gradient">Vote</h1>
...
           type="formfield"
-          click.delegate="deal.vote(true)"
+          click.delegate="acceptDeal()"
         >
-          <i class="fas fa-check-circle"></i>
-          ${deal.representativeVote(ethereumService.defaultAccountAddress) === true ? 'Accepted': 'Accept' }
+          <span class="icon is-loading" if.bind="accepting">
+            <i class="spinner fas fa-sync-alt" data-test="loadingIcon"></i>
+          </span>
+          <i class="fas fa-check-circle" if.bind="!accepting"></i>
```
dkent600, line 20: you can just use `else` here

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/pbutton/pbutton.scss
IonelLupu: yes. pbuttons don't add spaces for items inside it except when it's using the loading icon. In our case, we are using our own custom icons and thus they won't have any space between them and the text. 

With this change, any custom icon we will be using will have the needed space

dkent600 (approved):

PR #534, opened 2022-04-02
Fix bug #533

vipergtsrz:
Changed "container" to "pchip-container" on the pchip html and scss so it wouldn't bleed into other styles across the site


hiaux0 (approved):
back to normal
image

dkent600 (approved):

PR #539, opened 2022-04-04
added the "make an offer" section

IonelLupu:
## What was done
Polished the deal dashboard page
added the "make an offer" section


src/dealDashboard/dealMenubar/dealMenubar.ts
```diff
@@ -19,9 +19,14 @@ export class DealMenubar {
     this.eventAggregator.publish("showMessage", "Deal URL copied");
   }
 
-  @computedFrom("deal.status")
+  @computedFrom("deal.status", "deal.isCancelled")
   get canEdit() {
-    return !this.deal.isExecuted && !this.deal.isCancelled;
+    return this.deal.isUserProposalLead && (!this.deal.isCancelled || !this.deal.isExecuted);
```
dkent600, line 24: this tells me that it can be edited even if is executed, which is not correct, or if is cancelled, which is not correct.

dkent600 (approved):

PR #544, opened 2022-04-04
460- Like/Dislike should update immediately

giladt:
## What was done
Updated Like action to appear instantly in the UI. If API fails, action is reverted and a popup informs about the failure.


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -314,16 +314,41 @@ export class DiscussionThread {
...
       }
+
+      this.threadDictionary[_id] = {...message};
+      this.threadComments = Object.values(this.threadDictionary);
+
+      this.discussionsService.voteComment(this.discussionId, _id, type).then(
+        (success: boolean) => {
+          if (!success) {
+            /* On API failure- revert voting */
+            this.threadDictionary[_id] = {...swrVote};
```
hiaux0, line 344: Have you tested the failure case?
`swrVote` is just a shallow copy, so `upvotes` array property is still from local voting

dkent600 (approved):

Issue #547, opened 2022-04-04
[disclaimer] Width too small


tamagosante:
![image](https://user-images.githubusercontent.com/30693990/161550871-e53461f4-5422-4226-9bd6-d76607d54816.png)

curvelabs-dev:
Looks like leaking CSS to me.

PR #552, opened 2022-04-04
Fix bug #547 Add a class name to the popup modal for swapping

vipergtsrz:
Fixed the CSS bleeding in the funding page


hiaux0 (approved):
back to normal
(did not check the funding modal)

dkent600 (approved):

PR #565, opened 2022-04-05
Feat/proposal lead

vipergtsrz:
Change the funding page for the proposal lead view


src/funding/funding.html
```diff
@@ -40,62 +35,78 @@ <h1 class="title heading heading1 gradient">
...
-              </address-link>
-            </div>
           </div>
-          <div class="deposit">
-            <div
-              class="form"
-              if.to-view="!deal.isUserProposalLead && !deal.isTargetReached && !fundingFailed">
+          <div
+            class="deposit"
+            if.to-view="!deal.isUserProposalLead && !deal.isTargetReached && !fundingFailed"
```
hiaux0, line 55: Nit: Can we have separate vars to keep track of the state?
Eg. in line 39, we have the same check

src/funding/funding.ts
```diff
@@ -79,27 +78,29 @@ export class Funding {
...
+    this.secondDaoTokens.forEach(x => {this.setTokenContractInfo(x, this.secondDao);});
 
     //TODO figure out what the vested amount is from the deal
     //this.vestedAmount = this.deal.vestedAmount;
     this.vestedAmount = 0;
 
     //get contract token information from the DAO related to the account
-    this.daoRepresentedByCurrentAccountTokens = JSON.parse(JSON.stringify(this.deal.daoRepresentedByCurrentAccount.tokens));
-    this.daoRepresentedByCurrentAccountTokens.forEach(x => {this.setTokenContractInfo(x, this.deal.daoRepresentedByCurrentAccount);});
+    this.firstDaoTokens = JSON.parse(JSON.stringify(this.firstDao.tokens));
```
hiaux0, line 100: optional: this cloneDeep seems useful in general
--> Utils.cloneDeep

hiaux0 (commented):
all optional changes

src/funding/funding.ts
brandonseydel: @vipergtsrz - Check if the native JS function exists and use that else the above.

src/funding/funding.html
vipergtsrz: I added a new var for that state for you in a <let> tag.

src/funding/funding.ts
vipergtsrz: I added a new cloneDeep method and changed my code to call that. It uses the new built in structuredClone method if it's supported and if it's not it falls back to JSON.parse(JSON.stringify())

src/funding/deposit-grid/deposit-grid.scss
```diff
@@ -0,0 +1,103 @@
+@import "styles/styles.scss";
+@import "src/resources/elements/primeDesignSystem/styles";
+@import "src/resources/elements/primeDesignSystem/variables";
+@import "src/resources/elements/primeDesignSystem/colors";
+.deposit {
```
dkent600, line 5: I don't see this class being used by deposit-grid.  I would recommend calling it "deposit-grid" if it is the container of the component by that name.

src/funding/funding.ts
```diff
@@ -338,6 +339,7 @@ export class Funding {
    */
   private setTokenContractInfo(token: ITokenFunding, dao: IDAO): void {
     //get the additional token information from the contract for this token
+    console.log("Dao token transactions", this.deal.daoTokenTransactions.get(dao));
```
dkent600, line 342: will need to delete this eventually

dkent600 (approved):

src/funding/funding.ts
vipergtsrz: I'll make sure to delete this in the next PR, thanks!

PR #573, opened 2022-04-05
Fix/515 deletion of comments should not cause errors

giladt:
## What was done
If a user votes on a recently deleted comment, a popup appears saying that the "author has deleted that message", and it would be removed from the thread.

Fixed issue causing a deleted message, to reappear in the thread.


hiaux0:
515

(disconnected zenhub connection to 515, due to moving-on-the-board issues

PR #578, opened 2022-04-05
453 - Representative can only be part of one dao

hiaux0:
## What was done
- Add error message
image
- update related e2e tests to cover multiple reps (and clauses)

## Testing
1. prim dao rep address
2. partn dao same rep address
3. --> error message on blu

#### Before
was okay

#### After
error


hiaux0:
@BartuCulha 
Wdyt about the error message?

dkent600 (approved):

PR #585, opened 2022-04-06
deal dsahboard responsiveness

IonelLupu:
#### After

### DIFF
--- a/src/dealDashboard/dealDashboard.scss
+++ b/src/dealDashboard/dealDashboard.scss
@@ -12,7 +12,7 @@
     "clauses aside"
     "discussion discussion"
     "thread thread";
-  gap: 0 115px;
+  gap: 60px 115px;
 
   deal-menubar {
     grid-area: navbar;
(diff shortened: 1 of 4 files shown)


dkent600 (changes requested):
Need less left/right padding for clauses at smaller browser widths:

image

giladt:
Can you please add some margin-top above the discussions (I think it is relevant for all views...): 

image

src/dealDashboard/dealDashboard.scss
```diff
@@ -12,15 +12,14 @@
...
   deal-menubar {
     grid-area: navbar;
   }
 
   .top {
     grid-area: top;
-    margin-bottom: 40px;
 
     .title {
       padding-bottom: 12px;
```
giladt, line 25: If a title is too long to fit in view, I suggest breaking it.
```suggestion
      padding-bottom: 12px;
      word-break: break-all;
```

### Current:
image
### Expected:
image

giladt (changes requested):

dkent600 (approved):

PR #587, opened 2022-04-06
414 - Token details sub form wrong address

hiaux0:
## What was done
fix(tokenDetails): always reset token details 

Reason: 
- make sense upon address change to always reset

Background: 
- 414 still shows token details, despite wrong address

Cause:
- Code did not reset Token details on address guard

#### After
form always resets (and persists the reset) on stage change

### DIFF
--- a/src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
+++ b/src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
@@ -16,6 +16,7 @@ import { formatEther } from "ethers/lib/utils";
 import { NumberService } from "../../../../services/NumberService";
 import { PrimeRenderer } from "../../../../resources/elements/primeDesignSystem/validation/primeRenderer";
 import { WizardType } from "../../dealWizardTypes";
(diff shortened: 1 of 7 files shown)


src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
```diff
@@ -171,6 +178,8 @@ export class TokenDetails {
   }
 
   private addValidation() {
+    if (this.form.errors.length > 0) return;
```
IonelLupu, line 181: what is this condition for?

dkent600: would be good to comment this.

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
hiaux0: sorry, dev remnant. Pushed a remove commit

dkent600 (approved):

PR #595, opened 2022-04-06
fixed token details precision for amounts

IonelLupu:


src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
```diff
@@ -2,29 +2,26 @@ import { bindable } from "aurelia-typed-observable-plugin";
 import { bindingMode, computedFrom, customElement } from "aurelia-framework";
 import "./prange-slider.scss";
 import { Utils } from "../../../../services/utils";
-import { NumberService } from "../../../../services/NumberService";
-import tippy, { Instance } from "tippy.js";
+import { toBigNumberJs } from "../../../../services/BigNumberService";
+import { BigNumber } from "ethers";
```
hiaux0, line 6: Is there a way to remove these `BigNumber` packages from `<prange-slider>`?
Those seems too specific for a general p-system component

dkent600: I think is ok.  the component for now only supports bignumbers, which is all we need.

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
```diff
@@ -62,10 +54,8 @@ export class PRangeSlider {
       this.alreadyUpdated = false;
       return;
     }
-    this.left = this.numberService.toString(this.percentageToAbsoluteValue(100 - this.value), {mantissa: 0});
-    this.right = this.numberService.toString(this.percentageToAbsoluteValue(this.value), {mantissa: 0});
-
-    this.updateTooltips();
+    this.left = this.percentageToAbsoluteValue(Number((1 - this.value).toFixed(2)));
```
dkent600, line 57: we are supposed to be showing all the decimals.  That is why we got rid of the tooltips.

dkent600 (changes requested):
1) STEPS:

1. Go to Open Proposal wizard, Token Details
2. Type a number into Total Amount of Tokens to Swap

RESULT:
Crash

image

2) Should not be showing the ".0" here.  Should only show what the user entered.  Commas are fine.

image

src/resources/elements/primeDesignSystem/prange-slider/prange-slider.ts
IonelLupu: `.toFixed(2)` is not used to show the actual value shown in the UI. On this line `.toFixed` is used to limit the percentage value to 2 decimal places between 0.00 and 1.00. `Number((1 - this.value).toFixed(2)` was actually added because Javascript returns `0.45999999999999996` when doing `1 - 0.54` making the `percentageToAbsoluteValue` return an error because BigNumbers can't handle decimal places like that. So I had to use `.toFixed(2)` to transform it to `0.46`.

dkent600: Still, we should be showing all the decimals in the left and right input fields.  That is why we got rid of the tooltips.

dkent600 (approved):

PR #596, opened 2022-04-06
Implement deals observability

paweljanicki:
## What was done
After deals are initially loaded a subscription to deal updates is initiated.

**NOTE:** even though DealService is updated with new deals, the UI doesn't refresh automatically, so for example you have to go from all deals page to deal dashboard to actually see the updated deal

NOTE 2: Update to firebase functions: I've commented out code updating `modifiedAt` field because it was triggering multiple updates on the frontend pointlessly. That field is not used anywhere at the moment, but if we want it, we should look into a solution.

NOTE3: At the moment, deals in `this.deals` Map in `DealService` are updated more often than they should, because updates from firebase are emitted from 3 separate queries and sometimes deals are duplicated in those queries. Something to optimise. Suggestions welcomed

## Testing
- Adding new deals, changing deals registration data, votes.

#### Before
- Deals created/edited outside of the current browser tab, were not reflected in the app (without a refresh)

#### After
- Deals created/edited outside of the current browser tab are reflected in the UI

### DIFF
--- a/src/services/DataSourceDealsTypes.ts
+++ b/src/services/DataSourceDealsTypes.ts
@@ -1,8 +1,10 @@
+import { Observable } from "rxjs";
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import { IDealDiscussion } from "entities/DealDiscussions";
 import { IDealRegistrationTokenSwap } from "entities/DealRegistrationTokenSwap";
 import { IDealTokenSwapDocument } from "./../entities/IDealTypes";
 import { Address } from "./EthereumService";
+import { IDocumentUpdates } from "./FirestoreTypes";
 
 export type IDealIdType = string;
 
@@ -34,6 +36,11 @@ export abstract class IDataSourceDeals {
   getDeals<TDealDocument extends IDealTokenSwapDocument>(accountAddress?: Address): Promise<Array<TDealDocument>> {
     throw new Error("Method not implemented.");
   }
+
+  getDealsObservables(accountAddress?: Address): Promise<Observable<IDocumentUpdates<IDealTokenSwapDocument>>> {
+    throw new Error("Method not implemented.");
+  }
+
   /**
    * add new vote or update existing
    * @param dealId
--- a/src/services/DataSourceDealsTypes.ts
+++ b/src/services/DataSourceDealsTypes.ts
@@ -37,7 +37,7 @@ export abstract class IDataSourceDeals {
     throw new Error("Method not implemented.");
   }
 
-  getDealsObservables(accountAddress?: Address): Promise<Observable<IDocumentUpdates<IDealTokenSwapDocument>>> {
+  getDealsObservables(accountAddress: Address, skipFirst = false): Promise<Observable<IDocumentUpdates<IDealTokenSwapDocument>>> {
     throw new Error("Method not implemented.");
   }
 
--- a/src/services/DealService.ts
+++ b/src/services/DealService.ts
@@ -97,10 +97,11 @@ export class DealService {
         StartingBlockNumber = 0;
         break;
     }
-    this.eventAggregator.subscribe("Network.Changed.Account", (): void => {
+    this.eventAggregator.subscribe("Network.Changed.Account", async (): Promise<void> => {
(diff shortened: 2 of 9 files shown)


IonelLupu (approved):

vipergtsrz (approved):

src/services/DataSourceDealsTypes.ts
```diff
@@ -34,6 +36,11 @@ export abstract class IDataSourceDeals {
   getDeals<TDealDocument extends IDealTokenSwapDocument>(accountAddress?: Address): Promise<Array<TDealDocument>> {
     throw new Error("Method not implemented.");
   }
+
+  getDealsObservables(accountAddress?: Address): Promise<Observable<IDocumentUpdates<IDealTokenSwapDocument>>> {
```
dkent600, line 40: Please add code comments explaining the purpose of this function

src/services/DealService.ts
```diff
@@ -108,7 +110,52 @@ export class DealService {
     /**
      * deals will take care of themselves on account changes
      */
-    this.getDeals();
+    if (!this.deals) {
```
dkent600, line 113: initialize is only called on app startup (in main).  `!this.deals` should always be true.  can we delete this or comment on why it is necessary?

src/services/DealService.ts
```diff
@@ -108,7 +110,52 @@ export class DealService {
...
+      await this.getDeals();
+    }
+
+    this.observeDeals();
+  }
+
+  private async observeDeals(): Promise<void> {
+    this.deals = new Map<Address, DealTokenSwap>();
```
dkent600, line 121: do you need to set `this.initializing = true` like getDeals does?  How do you handle reentrancy?

src/services/DealService.ts
```diff
@@ -58,6 +59,7 @@ export class DealService {
 
   public initializing = true;
   private initializedPromise: Promise<void>;
+  private dealsSubscription: Subscription;
```
dkent600, line 62: Can you please comment what is the purpose of this?

src/services/DealService.ts
dkent600, line 121: You should not set this.deals until the Map is completely populated.  Use a local Map and assign at the end, like getDeals does.  This will help minimize thrashing.

src/services/DealService.ts
```diff
@@ -108,7 +110,52 @@ export class DealService {
...
+          for (const dealDoc of updates.removed) {
+            this.deals.delete(dealDoc.id);
+          }
+
+          for (const dealDoc of updates.modified) {
+            this._createDeal(dealDoc, dealsMap);
+          }
+
+          dealsMap.forEach(deal => {
+            this.deals.set(deal.id, deal);
```
dkent600, line 148: why do this?  isn't it already done above?

src/services/DealService.ts
```diff
@@ -108,7 +110,52 @@ export class DealService {
...
+          }
+
+          const dealsMap = new Map<Address, DealTokenSwap>();
+
+          for (const dealDoc of updates.removed) {
+            this.deals.delete(dealDoc.id);
+          }
+
+          for (const dealDoc of updates.modified) {
+            this._createDeal(dealDoc, dealsMap);
```
dkent600, line 144: If only modified, why are you creating a new entity?  Don't we simply want to replace the  existing deal.dealDocument?

dkent600 (changes requested):

src/services/DealService.ts
```diff
@@ -182,6 +193,55 @@ export class DealService {
     return dealIds;
   }
 
+  private async observeDeals(): Promise<void> {
+    const dealsObservable = await this.dataSourceDeals.getDealsObservables(this.ethereumService.defaultAccountAddress, true);
```
paweljanicki, line 197: By passing `true` as the second argument we skip first emitions of all queries. In my opinion we should always skip the first, because `observeDeals` method is always triggered after `getDeals`, so the deals are "fresh".

If we would only skip first query emitions when triggering  `observeDeals` for the first time, then it wouldn't skip first on accountAddress change.

PR #597, opened 2022-04-06
Remove postinstall script which was breaking vercel deployment

paweljanicki:
I leave it up to @brandonseydel to decide if we have to merge this


paweljanicki:
Not necessary

PR #607, opened 2022-04-07
593 - Catch thread comments undefined

hiaux0:
Related
599
601
603

## What was done
- throw error for network call in `DiscussionService`
  - Why? So caller can catch and handle api related errors (eg. show retry)
- Early return `ensureDealDiscussion`

## Testing
Right now, unstable, because convo sometimes okay, sometimes not

#### Before
always infinite loading

#### After
handle, and display default "This discussion has no comments yet."

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lint.fix": "npm run lint -- --fix",
     "fetchContracts": "node scripts/fetchContracts.js",
     "firebase": "npm run firebase-functions:build && npm run firebase-emulators:start",
-    "firebase-e2e": "npm run firebase-functions:build && firebase emulators:start --project=default --import=firebase_seed_e2e",
+    "firebase-e2e": "firebase emulators:start --project=default --import=firebase_seed_e2e",
     "firebase:all": "npm run firebase-functions:watch & npm run firebase-emulators:start",
     "firebase-update-seed-data": "firebase emulators:export --force firebase_seed",
     "firebase-update-seed-data:e2e": "firebase emulators:export --project=default --force firebase_seed_e2e",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lint.fix": "npm run lint -- --fix",
     "fetchContracts": "node scripts/fetchContracts.js",
     "firebase": "npm run firebase-functions:build && npm run firebase-emulators:start",
-    "firebase-e2e": "firebase emulators:start --project=default --import=firebase_seed_e2e",
+    "firebase-e2e": "npm run firebase-functions:build && firebase emulators:start --project=default --import=firebase_seed_e2e",
     "firebase:all": "npm run firebase-functions:watch & npm run firebase-emulators:start",
     "firebase-update-seed-data": "firebase emulators:export --force firebase_seed",
     "firebase-update-seed-data:e2e": "firebase emulators:export --project=default --force firebase_seed_e2e",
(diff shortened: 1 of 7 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -205,40 +226,38 @@ export class DiscussionThread {
...
-      this.updateDiscussionListStatus(
-        new Date(parseFloat(this.threadComments[this.threadComments.length - 1].createdOn)),
-        this.threadComments.length,
-      );
-    }
+    // Update the discussion status
+    this.updateDiscussionListStatus(
+      new Date(parseFloat(this.threadComments[this.threadComments.length - 1].createdOn)),
+      this.threadComments.length,
+    );
```
hiaux0, line 260: This whole block, I just removed the surrounding if statement `    if (this.threadComments && this.dealDiscussion) {` (see red diff line 208)

Reason: Moved the if check earlier

dkent600 (approved):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -193,7 +200,21 @@ export class DiscussionThread {
...
-    this.threadComments = await this.discussionsService.loadDiscussionComments(discussionId);
+    try {
+      this.threadComments = await this.discussionsService.loadDiscussionComments(discussionId);
+
+      // Early return, if there are no comments/discussions.
+      if (!this.threadComments) return;
+
+      this.apiErrorText = "";
+    } catch (error) {
+      this.apiErrorText = "A problem occured while loading the discussion";
```
dkent600, line 211: can this text be in the HTML instead of the viewmodel?

package.json
```diff
@@ -27,7 +27,7 @@
     "lint.fix": "npm run lint -- --fix",
     "fetchContracts": "node scripts/fetchContracts.js",
     "firebase": "npm run firebase-functions:build && npm run firebase-emulators:start",
-    "firebase-e2e": "firebase emulators:start --project=default --import=firebase_seed_e2e",
+    "firebase-e2e": "npm run firebase-functions:build && firebase emulators:start --project=default --import=firebase_seed_e2e",
```
dkent600, line 30: shouldn't be in this PR

dkent600 (changes requested):

hiaux0:
#### Fix Verification
- Went through all related tickets and made sure, that described issues are fixed by this
  -  --> yes, because all of them write about `cannot read undefined` or are network related
- additional test case, that mocks network error _once_, then a "Reload Discussions" actually shows, in this case, "No discsussions" text 
  - recording of test case:

reload-discussions-api-error

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -86,7 +89,7 @@ export class DiscussionThread {
     return (discussionsIds.indexOf(this.discussionId) + 1).toString() || "-";
   }
 
-  @computedFrom("isLoading.discussions", "deal.isUserRepresentativeOrLead", "threadComments")
+  @computedFrom("isLoading.discussions", "deal.isUserRepresentativeOrLead", "threadComments", "apiErrorText")
```
dkent600, line 92: does apiErrorText really exist?

Can we move the actual text copy into the HTML?

dkent600 (changes requested):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
hiaux0: argh.. is remnant change.

(next time will also search for variable, instead of just renaming command)

dkent600 (approved):
Would like to move the text copy on lines 96 and 97 of discussionThread.ts into the HTML

PR #608, opened 2022-04-07
Handle wallet change, while deals are initialising for another wallet

paweljanicki:
# Error when changing account in MetaMask:
**Background:**
Representative addresses and proposal lead address are stored as is, they are NOT transformed to lowerCase.
All queries also don’t transform addresses to lowerCase, e.g. where("registrationData.proposalLead.address", "==", address) 
When user is authenticated to Firebase, we also don’t transform the address to lowerCase.

**Issue:**
Somehow, when using MM and Chrome, this event in FirebaseService, emits address transformed to lowerCase, when changing account in MM

  public initialize(): void {
    this.eventAggregator.subscribe("Network.Changed.Account", (address: Address) => {
      this.syncFirebaseAuthentication(address);
    });
  }

When refreshing the page, first time it emits address that is NOT lowerCase.
But everytime the address is changed in MM it emits lowercase address.

Therefore, the user was authenticated to Firebase with lowercase address, and that why there was permissions issue, because firestore rules, as well as queries where looking for NOT lowerCase address but they were getting a lowerCase one.

**Solution:**
Always convert addresses to lowerCase

## TODO
Update all data in all firebase project, to have all addresses stored in lowerCase.
That includes updating vote documents in primary-dao-votes and partner-dao-votes subcollections, to have document id (equals to wallet address) in lowerCase

# Issue when changing wallet twice quickly (while deals are still initialising for the previous address)

## What was done
Changing wallet while deals are being initialized for another wallet, causes getting deals from dataSourceDeals to fail. When it fails in my opinion we should reset deals Map in DealService.

GetDeals method in DealService should we triggered every time an address is changes, so for example:
1. I'm connected to wallet A and I can see deals for wallet A
2. I switch to wallet B, getDeals method is triggered, but before it can finish I switch back to wallet A
3. Getting deals for wallet B fails, because wallet B is no longer authenticated with firebase
4. getDeals method that was triggered for wallet B, resets deals Map (because it failed to get deals, it shouldn't show the old ones)
5. (not implemented) After switching back to wallet A, getDeals method should be triggered again, then when it succeeds, deals Map should be set to the deals for wallet A, coming from getDeals method. It doesn't work because currently on account address change, we change if initializing is in progress.
```
this.eventAggregator.subscribe("Network.Changed.Account", async (): Promise<void> => {
    // because of this condition, getDeals is not triggered when other getDeals is still in progress (initializing),
    // therefore we cannot correctly react to account changes
      if (!this.initializing) { 
        try {
          this.eventAggregator.publish("deals.loading", true);
          await this.getDeals(true);
        } catch {
          this.eventAggregator.publish("deals.loading", false);
        } finally {
          this.eventAggregator.publish("deals.loading", false);
        }
      }
    });
```

## Testing
Start app with wallet account A. Open wallet and change the account to B and rapidly back to A, (you might have up to a few seconds for this). It might be difficult to reproduce this when using firebase emulators, when debugging locally, connect to live firebase project.

#### Before
App was broken when wallet change occurred while deals where being initialised for another wallet.
In other words the app was try to load private deals for a user, and the user got changed. So the first user was no longer authenticated to firebase, yet the app was still trying to get the deals for them, chance the error "Missing or insufficient permissions." That error wasn't handled.

#### After
Handle error while loading deals for a user, reset deals Map

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,3 +1,4 @@
+import { DealService } from "services/DealService";
 import "./styles/styles.scss";
 import "./app.scss";
 
@@ -11,7 +12,6 @@ import { ConsoleLogService } from "services/ConsoleLogService";
 import { EthereumService } from "services/EthereumService";
 import { EventAggregator } from "aurelia-event-aggregator";
 import { EventConfigException } from "services/GeneralEvents";
-import { FirebaseService } from "./services/FirebaseService";
 import { PLATFORM } from "aurelia-pal";
 import { ShowButtonsEnum } from "resources/elements/primeDesignSystem/ppopup-modal/ppopup-modal";
 /* eslint-disable linebreak-style */
@@ -29,7 +29,7 @@ export class App {
     private consoleLogService: ConsoleLogService,
     private alertService: AlertService,
     private storageService: BrowserStorageService,
-    private firebaseService: FirebaseService) { }
+    private dealService: DealService) { }
 
   router: Router;
   onOff = false;
@@ -45,7 +45,7 @@ export class App {
     return false;
   };
 
-  public attached(): void {
+  public async attached(): Promise<void> {
     // so all elements with data-tippy-content will automatically have a tooltip
     tippy("[data-tippy-content]");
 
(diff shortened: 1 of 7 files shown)


dkent600:
STEPS:

1. start the app, wait until the deals have loaded
2. connect app to a wallet
3. in MM, change the wallet

RESULT:
App shows infinite spinny icon

dkent600 (approved):

PR #614, opened 2022-04-07
414 - Token details not revalidating 

hiaux0:
## What was done

Round 3 for ticket 414.
I tested all previous cases manually, and those also have e2e tests.
Waiting for the next bug report to come..

#### Before
no validation

#### After
validation

### DIFF
--- a/src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
+++ b/src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
@@ -188,7 +188,6 @@ export class TokenDetails {
       .ensure<IToken, string>(data => data.address)
       .required()
       .satisfiesRule(Validation.isETHAddress)
-      .then()
       .satisfiesRule("isValidIERC20Address")
       .ensure<string>(data => data.amount)
       .required()
--- a/cypress/integration/common/deal-common.ts
+++ b/cypress/integration/common/deal-common.ts
@@ -93,7 +93,7 @@ Given("I create a Private Partnered Deal", () => {
 });
 
 Given("I create an Open Proposal", () => {
-  const deal = DealDataBuilder.create().withProposalLeadData({address: E2eWallet.currentWalletAddress}).deal;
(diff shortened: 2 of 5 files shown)


src/wizards/tokenSwapDealWizard/components/tokenDetails/tokenDetails.ts
```diff
@@ -188,7 +188,6 @@ export class TokenDetails {
       .ensure<IToken, string>(data => data.address)
       .required()
       .satisfiesRule(Validation.isETHAddress)
-      .then()
```
hiaux0, line 191:
```
Validate subsequent rules after previously declared rules have been validated successfully. 
Use to postpone validation of costly rules until less expensive rules pass validation.
```
Caused the following ERC20 rule not to trigger, thus the back.

---

The type of one-line fixes, that cause you hours of debugging :)

IonelLupu:
I did some tests on the "development" branch (that has the `.then`).

It seems that clicking "Save" doesn't run the `isValidIERC20Address` validator. Looks like this validator runs only when you type something in the input. Here are some images that show that:
image
In the above image I pasted the address from the e2e tests in the Address input. As shown by the breakpoint, it returns a "result" for the `isValidIERC20Address` validator. This is as expected behavior because we want to see that error message.

But, in the below image I pressed the "Save" button when having the same address in the input. Notice there are only two results for the `address` input. There is no `isValidIERC20Address` with `valid: false` in the results list. This is unexpected behavior. I was expecting it validates all of the rules after `.then` when the rules before `.then` are all valid. It seems it doesn't do that.

image



We might need to revise the docs for Aurelia validator.

The issue with removing `.then` is that the user will have to wait a few seconds to see the error message. If you type "asd" in the address input, after 1s or so, you will see the "Please enter a valid ethereum address" error message. If you empty out the input, after 1s or so, you will see the "Address is required." error message. 

I just wanted to let you know why `.then` was added there. It fixes the issue described in the tests but breaks the UX a bit.

hiaux0:
@IonelLupu Thanks for the detailed analysis. Feel free take over the ticket.

I'm okay with the UX tradeoff for a working form + validation.
Of course I'd prefer better ux as well.
I suggest optional improvement

IonelLupu:
@hiaux0  maybe @brandonseydel has some ideas here

dkent600 (approved):

PR #620, opened 2022-04-08
styles(design.input): add label margin

hiaux0:
## What was done
(quick fix, no ticket okay?)
- Mocks have bigger margin-bottom 
- fix is save, because only TokenDetails uses `label-description`

#### Before
image

#### After
image

### DIFF
--- a/src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
+++ b/src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
@@ -34,6 +34,7 @@ pform-input {
     white-space: pre-line;
     line-height: 24px;
     color: $Neutral03;
+    margin-bottom: 10px;
   }
 
   .inputContainer {
--- a/src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
+++ b/src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
@@ -34,7 +34,6 @@ pform-input {
     white-space: pre-line;
     line-height: 24px;
     color: $Neutral03;
-    margin-bottom: 10px;
   }
 
   .inputContainer {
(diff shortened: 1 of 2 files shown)


src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
```diff
@@ -34,6 +34,7 @@ pform-input {
     white-space: pre-line;
     line-height: 24px;
     color: $Neutral03;
+    margin-bottom: 10px;
```
dkent600, line 37: this makes me nervous that it can affect a lot of places.  Have you checked them all?

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
hiaux0: > fix is save, because only TokenDetails uses label-description

Yup.
I did a search for `label-description`

dkent600: To be safe, since only token details uses label-description, lets move this into there.  I have been dealing with issues with pform-input (when the label and input are horizontally oriented) and really prefer not to mess with this component.  Is too fragile ATM.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
hiaux0: lcoalized to token-details


[styles(tokenDetails): add bottom margin to label description](https://github.com/PrimeDAO/prime-deals-dapp/pull/620/commits/ee7464145066f62e5b3925040d0c9cbaece9e373)

PR #624, opened 2022-04-08
added a test for a bug that shows up when changing the address

IonelLupu:
## What was done
added a test for a bug that shows up when changing the address from metamask and the deals don't load anymore

#### After

### DIFF
--- a/cypress/integration/common/index.ts
+++ b/cypress/integration/common/index.ts
@@ -1,0 +1,10 @@
+import { And } from "@badeball/cypress-cucumber-preprocessor/methods";
+
+And("I Wait for the modal with the message {string} to disappear", (modalContent: string) => {
+  cy.contains("[data-test='modelContent']", modalContent).should("be.visible");
+  cy.get("[data-test='modelContent']").should("not.be.visible");
+});
+
+And("The modal with the message {string} is hidden", (modalContent: string) => {
+  cy.contains("[data-test='modelContent']", modalContent).should("not.be.visible");
+});
--- a/cypress/integration/features/deals/all-deals/view-all-deals.feature
+++ b/cypress/integration/features/deals/all-deals/view-all-deals.feature
@@ -12,3 +12,13 @@ Feature: View All Deals
     Scenario: View partnered deals
         When I select Partnered Deals tab
         Then I can see Partnered Deals
+
+    @focus
+    Scenario: User can successfully change Metamask accounts
+        Given I'm not connected to a wallet
+        And I connect to the wallet with address "0x0000000000000000000000000000000000000000"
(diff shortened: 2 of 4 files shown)


dkent600 (approved):

cypress/integration/features/deals/all-deals/view-all-deals.feature
```diff
@@ -12,3 +12,13 @@ Feature: View All Deals
...
+
+    @focus
+    Scenario: User can successfully change Metamask accounts
+        Given I'm not connected to a wallet
+        And I connect to the wallet with address "0x0000000000000000000000000000000000000000"
+        And I Wait for the modal with the message "Thank you for your patience while we initialize for a few moments..." to disappear
+        And I change the address to "0x0000000000000000000000000000000000000001"
+        And I Wait for the modal with the message "Thank you for your patience while we initialize for a few moments..." to disappear
+        Then The modal with the message "Thank you for your patience while we initialize for a few moments..." is hidden
+        And I'm connected to my wallet with address "0x0000000000000000000000000000000000000001"
```
hiaux0, line 24: 3 Notes

1. Metamask is not our app, so imo, we shouldn't have this as a Scenario, which relates to our app...
2. ... For this, I suggest to add either another gherkin tag (eg. "@thirdParty"), or use existing @regression
3. I would have written this differently, as the current way is too explicit about the UI:
```feature
Scenario: ...
  Given ...
  And I successfully connect to the wallet ...
  When my Wallet gets changed by a 3rd party 
  Then I'm connected to the new address
```

This way, we might be able to address 1. as well.
Optionally, hard-coding the address is fine for illustration purposes, but would have been better to use an more obvious example

hiaux0: Also, as I understand it, was this bug fixed?
Or will it break currently?

cypress/integration/common/index.ts
```diff
@@ -0,0 +1,10 @@
+import { And } from "@badeball/cypress-cucumber-preprocessor/methods";
```
hiaux0, line 1: would have been nice to at least put in an existing file, so not another file was yet introduced.
`wallet.e2e.ts` also contains "modelContent", or else, there is just `deals.e2e.ts`

IonelLupu: I added these steps in the `index.ts` file because they look rather general. A modal can be displayed on any page for any purpose. I thought we can use these in the future for some other tests.

cypress/integration/features/deals/all-deals/view-all-deals.feature
IonelLupu: I did some more tests and it seems, even if the bug is fixed, the test is not quite correct. I shall modify it

PR #631, opened 2022-04-10
fixed dancing privacy button

IonelLupu:
## What was done
fixed dancing privacy button
#### After


dkent600 (changes requested):
If for some reason deal.setPrivacy is being re-entered, then it should be handled in deal.setPrivacy, not here.  Or better yet, should be handled in the toggle button or the component that contains it.  I am curious what is the root cause of this reentrancy?

Further, the message should be published not here but from the component that contains the toggle button (dealInfo).

src/dealDashboard/dealInfo/dealInfo.ts
```diff
@@ -2,12 +2,47 @@ import { autoinject, bindable } from "aurelia-framework";
...
+  ) {
+  }
+
   getSocialMediaDetails(socialMediaName: string) {
     return availableSocialMedias.find(socialMedia => socialMedia.name === socialMediaName);
   }
+
+  public async attached() {
+    this.subscriptions.push(
+      this.aureliaHelperService.createPropertyWatch(this.deal, "isPrivate", async (value: boolean) => {
```
dkent600, line 28: Can we remove this subscription and either:

1) have a method in this viewmodel that is executed on click of the toggle and does therein whatever is needed, or better:
2) have the toggle support a bindable click method that is invoked on click and does therein whatever is needed.

I vote #2.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/ptoggle/ptoggle.html
```diff
@@ -6,6 +6,7 @@
       value.bind="value"
       class="pToggleInput ${ validationState ? 'pInput-' + validationState : '' }"
       disabled.bind="disabled"
+      change.delegate="changed()"
```
dkent600, line 9: This should not be made to look or behave like an event.  It should look and behave like an optional bound property whose type is `(checked: boolean) => void` that the ptoggle viewmodel calls when the component is clicked.

dkent600 (changes requested):

src/resources/elements/primeDesignSystem/ptoggle/ptoggle.html
```diff
@@ -6,7 +6,7 @@
       value.bind="value"
       class="pToggleInput ${ validationState ? 'pInput-' + validationState : '' }"
       disabled.bind="disabled"
-      change.delegate="changed()"
+      change.delegate="changed({checked: value})"
```
dkent600, line 9: what happens if changed isn't supplied?

dkent600 (changes requested):

dkent600 (approved):

PR #645, opened 2022-04-11
634 - Fix deleting/adding comments error

hiaux0:
- [x] wait until convo back up (12am 12.04.2022)

## What was done
Background:
- Removing prop directly from object - via `delete` keyword
- Caused issues in Aurelia Template

## Testing
1. Have 2 comments in the Thread
2. reload page
3. Add a new comment
4. delete any 2 comments

#### Before
5. broken ui

#### After
5. ok

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -123,7 +123,6 @@ export class DiscussionThread {
 
   private arrayToDictionary(comments: Array<any>): Record<string, IComment> {
     return comments.reduce((r, e): Record<string, IComment> => {
-      console.log("just e: ", e);
       if (e === undefined) {
         return r;
       }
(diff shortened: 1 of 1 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -145,18 +145,21 @@ export class DiscussionThread {
...
-        newComment.text;
-
+      const decrypted = await this.discussionsService.decryptWithAES(
+        newComment.metadata.encrypted,
+        newComment.metadata.iv,
+        key,
+      );
+      if (newComment.metadata.encrypted) {
+        newComment.text = decrypted;
+      }
```
hiaux0, line 162: Changed this block, because after updating the code, an eslint error appeared (sth with race-condition).

New way: _always_ decrypts, and _then_ checks if we have encrypted content.
Old way: only decrypt _after_ encyrpt check

Imo: Nothing changes, as we decrypt everything anyway.

hiaux0 (commented):
.

dkent600 (approved):

PR #650, opened 2022-04-12
Add token swap fee information to deal type selector

paweljanicki:
## What was done
- add fee info

## Testing
- open initiate a deal page
Screen Shot 2022-04-12 at 11 15 21 pm

#### After

### DIFF
--- a/src/initiate/initiate.ts
+++ b/src/initiate/initiate.ts
@@ -1,4 +1,5 @@
 import { IDealTypeBox } from "./dealTypeSelector/dealTypeSelector";
+import "./initiate.scss";
 
 export class Initiate {
   boxes: IDealTypeBox[] = [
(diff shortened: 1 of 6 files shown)


src/initiate/initiate.ts
```diff
@@ -7,6 +8,12 @@ export class Initiate {
       slug: "initiate/token-swap",
       isDisabled: false,
       description: "Safely exchange tokens between multiple parties with customized vesting schedules.",
+      specialContent: `
```
dkent600, line 11: can we please use a slot so we don't have to put html and text copy in the viewmodel?

dkent600 (changes requested):

src/initiate/initiate.ts
dkent600: this is actually another argument to ditch the javascript-generated boxes entirely.

dkent600:
Worked elsewhere

PR #656, opened 2022-04-13
feat(firebase) - Allow refresh of the data on the UI and via script t…

brandonseydel:
…through node

## What was done
- add reset button to navbar
- add 9 canonical deals 
  - 2 open prop (one minimal, one FlorestaDAO)
  - 7 part deal (minimal, private, FlorestaDAOxClimatiDAO, 4 deals to test funding)
    - ! Note: 
      - the funding deals always reset their ids, whereas the others keep their ids
      - and they should have all votes green (even after reset)

## Testing
- Click on "Reset Deals"
  - should always reset deals to canonical state
- also checkout emulators how the reset behaves (should reset as expected)
- voting works for funding deals (indicated by naming scheme " [2] Isnt:1 - VFor:1 - COf:0.5 ")

## TODO
- [x] @brandonseydel After resetting deals, sub-collections of votes from partnered deals are orphaned. They should be deleted, to avoid any confusion or unexpected bugs
- [x] @brandonseydel  scripts not working? (seed-run)
- [x] @brandonseydel @paweljanicki Remove createCustomToken
- [x] @paweljanicki Can you please review this? There are significant firebase-related changes, including what appears to be allowing anyone to change the vote of a representative.
- [x] @dkent600 @paweljanicki  Overall the script logic is not super clean (eg. copy pasted code, that was then also modified), imo, in a follow up, we should aim to clean this up


firestore.rules
```diff
@@ -24,7 +24,7 @@ service cloud.firestore {
     match /deals-token-swap/{deal} {
       allow read: if isPublicDeal() || userIsProposalLead() || userIsRepresentative();
       allow update: if userIsProposalLead() || (userIsAuthenticated() && updateOnlyToDiscussions());
-      allow create, delete: if false;
+      allow create, delete: if true;
```
hiaux0, line 27: Does this mean, anyone will be able to create?

package.json
```diff
@@ -10,6 +10,8 @@
   "homepage": "./",
   "license": "MIT",
   "scripts": {
+    "build-seed": "cd ./src/server-browser-scripts && tsc",
+    "run-seed": "node -e 'import(\"./dist/seed/src/server-browser-scripts/seed-data.js\").then(x=> x.resetDeals())' -r dotenv/config . dotenv_config_path=./.env",
```
hiaux0, line 14: `dotenv/config . dotenv_config_path=./.env`

not possible to use `cross-env DOTENV_CONFIG_PATH` like the other scripts here?

hiaux0 (changes requested):
## Getting build errors 

```
[tsl] ERROR in /home/hdn/dev/prime/prime-deals-dapp/src/services/FirestoreDealsService.ts(2,10)
      TS2459: Module '"./FirebaseService"' declares 'firebaseAuth' locally, but it is not exported.
```

## Fnality
Maybe related to above, but clicking on that button in local env
(npm run start-dev)

image

firestore.rules
brandonseydel: I wanted to ask to @BartuCulha about this.  Not sure we really care about this rule in anything other than prod.

package.json
brandonseydel: I believe I tried it and it doesnt work without all the above.

src/server-browser-scripts/seed-data.ts
```diff
@@ -0,0 +1,61 @@
...
+  collectionKey: string,
+  data: Record<string, unknown>,
+) => {
+  const { dealId, ...correctedData } = data;
+  return dealId ?
+    setDoc(doc(firebaseDatabase, collectionKey, String(dealId)), correctedData) :
+    addDoc(collection(firebaseDatabase, collectionKey), correctedData);
+};
+
+signInToFirebase("0xE834627cDE2dC8F55Fe4a26741D3e91527A8a498");
```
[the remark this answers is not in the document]
brandonseydel: Yea that needs to listen for wallet but default probably

hiaux0 (changes requested):
Builds now, but I'm not getting the expected JSON structure in Firestore

1. `default` should be `registrationData`
2. `registrationData` props are in the root (they should not)

image

firestore.rules
hiaux0: cc @paweljanicki

hiaux0:
Have an issue with the xyz-dao-votes document.

STEPS
2. look at a deal, that keeps their id, eg. Minimal_Partnered_deal
3. vote with both addresses
4. reload browser

RESULT
votes gone

EXPECTED 
Votes persist

---
I don't fully understand how voting is implemented, so appreciate some help.
What I see, when resetting, it still has documents 
image

cc @paweljanicki

dkent600 (changes requested):
1)  @paweljanicki Can you please review this?  There are significant firebase-related changes, including what appears to be allowing anyone to change the vote of a representative.

2) can we add a readme.md that explains usage of the scripts added to package.json, how one would go about changing the test data, and any other important information?

paweljanicki:
Screen Shot 2022-04-29 at 1 30 15 pm

After resetting deals, sub-collections of votes from partnered deals are orphaned. They should be deleted, just like sub-collections of open proposals

paweljanicki:
> Have an issue with the xyz-dao-votes document.
> 
> STEPS 2. look at a deal, that keeps their id, eg. Minimal_Partnered_deal 3. vote with both addresses 4. reload browser
> 
> RESULT votes gone
> 
> EXPECTED Votes persist
> 
> I don't fully understand how voting is implemented, so appreciate some help. What I see, when resetting, it still has documents image
> 
> cc @paweljanicki
> 
> @hiaux0: This happens, because resetting deals, doesn't create sub-collections for votes. I will suggest a fix for that



> Have an issue with the xyz-dao-votes document.
> 
> STEPS 2. look at a deal, that keeps their id, eg. Minimal_Partnered_deal 3. vote with both addresses 4. reload browser
> 
> RESULT votes gone
> 
> EXPECTED Votes persist
> 
> I don't fully understand how voting is implemented, so appreciate some help. What I see, when resetting, it still has documents image
> 
> cc @paweljanicki

@hiaux0: This happens, because resetting deals, doesn't create sub-collections for votes. I will suggest a fix for that

brandonseydel:
@paweljanicki - We need to meet to figure out the create custom token func and how I can use it without all the signer stuff. (only for node)

paweljanicki:
I've added changes to only deploy certain Firebase functions in the CI and to use different Firestore rules for production

paweljanicki (approved):
My only concern is that clicking "reset deals" button will sign in to Firebase with the testing wallet and will sign out the current user, but the connected wallet will remain the same and the app might behave weird.

brandonseydel:
> My only concern is that clicking "reset deals" button will sign in to Firebase with the testing wallet and will sign out the current user, but the connected wallet will remain the same and the app will behave weird - UI will be broken in a few places. I we make it a known issue, that's OK, otherwise we can expect incoming bugs

I can change this if this is not the intended behavior

package.json
hiaux0, line 14: I would prefer to use the other naming ef9a5a15861de78e24690b4d7ae98d1d6975e221 
`seed-build`
`seed-run`
and to put these scripts further down

if not, can you update the README.md?

src/server-browser-scripts/seed-data.ts
```diff
@@ -0,0 +1,224 @@
...
+ * @param collectionKey
+ */
+export const clearCollection = async (
+  collectionKey: string,
+) => {
+  const collectionRef = collection(firebaseDatabase, collectionKey);
+  const queryResult = await getDocs(collectionRef);
+  await Promise.all(queryResult.docs.map(async doc => {
+    const subCollection1 = collection(firebaseDatabase, collectionKey, doc.id, PRIMARY_DAO_VOTES_COLLECTION);
+    const subCollection2 = collection(firebaseDatabase, collectionKey, doc.id, PRIMARY_DAO_VOTES_COLLECTION);
```
hiaux0, line 46:
```suggestion
    const subCollection2 = collection(firebaseDatabase, collectionKey, doc.id, PARTNER_DAO_VOTES_COLLECTION);
```

image

hiaux0 (changes requested):

paweljanicki:
Ready to merge in my opinion

hiaux0 (approved):
- voting works
- resetting works
- connected account not affected

:green_apple:

PR #657, opened 2022-04-13
Observe deals updates in firebase and update only parts of dealDocument that were actually changed

paweljanicki:
## What was done
- Add deep-diff npm package
- Ignore document deletions from queries to firestore
- Update only parts of dealDocument that were changed

## Testing
- Update/observe deal on multiple browsers

#### After


src/services/DealService.ts
```diff
@@ -340,4 +337,22 @@ export class DealService {
...
+  /**
+   * Updates only parts of dealDocument that were updated
+   */
+  private updateDealDocument(dealDocument: IDealTokenSwapDocument, updatedDocument: IDealTokenSwapDocument):void {
+    // ignore updates to modifiedAt property
+    delete dealDocument.modifiedAt;
+    delete updatedDocument.modifiedAt;
+
+    // delete partnerDAO if undefined
+    if (!dealDocument.registrationData.partnerDAO) {
```
dkent600, line 350: should this be part of the microtask?

dkent600 (changes requested):

PR #667, opened 2022-04-14
Handle deals that were removed from observed firestore query

paweljanicki:
## What was done
Handle deals removed from a query and delete them from deals map if the user is not representative/PL

#### After

### DIFF
--- a/src/services/DataSourceDealsTypes.ts
+++ b/src/services/DataSourceDealsTypes.ts
@@ -4,6 +4,7 @@ import { IDealDiscussion } from "entities/DealDiscussions";
 import { IDealRegistrationTokenSwap } from "entities/DealRegistrationTokenSwap";
 import { IDealTokenSwapDocument } from "./../entities/IDealTypes";
 import { Address } from "./EthereumService";
+import { IDocumentUpdates } from "./FirestoreTypes";
 
 export type IDealIdType = string;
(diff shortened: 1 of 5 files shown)


src/services/DataSourceDealsTypes.ts
```diff
@@ -4,6 +4,7 @@ import { IDealDiscussion } from "entities/DealDiscussions";
 import { IDealRegistrationTokenSwap } from "entities/DealRegistrationTokenSwap";
 import { IDealTokenSwapDocument } from "./../entities/IDealTypes";
 import { Address } from "./EthereumService";
+import { IDocumentUpdates } from "./FirestoreTypes";
```
hiaux0, line 7: `DataSourceDealsTypes.ts` is supposed to be a generic interface.
Is there another way to specify this type? (and not add a dependency on firestore)

dkent600: agreed

dkent600: IDocumentUpdates  should be defined here, not in FirestoreTypes

dkent600 (changes requested):

paweljanicki:
Decision was made to refactor the code in a way that those changes are not relevant anymore

PR #671, opened 2022-04-15
showing the tooltip only when the dao name doesn't fit on the screen

IonelLupu:
#### After


dkent600:
Closing as we won't work the ellipses-with-tooltip feture request at this time.  Note that DAO names are rarely long.  The UI should accommodate the typical lengths without cropping.

PR #681, opened 2022-04-16
swapped order of deal info dao name labels

IonelLupu:
## What was done
image
image

#### After

### DIFF
--- a/src/dealDashboard/dealInfo/dealInfo.html
+++ b/src/dealDashboard/dealInfo/dealInfo.html
@@ -5,7 +5,7 @@
     <hr>
 
     <div class="infoSection">
-      <span class="bodySmallTextBold infoTitle">Primary<br>DAO</span>
+      <span class="bodySmallTextBold infoTitle">Primary DAO</span>
       <div class="daoDetails" ptooltip.bind="deal.primaryDao.name">
         <img class="daoLogo" src.bind="deal.primaryDao.logoURI">
         <div class="daoTitle">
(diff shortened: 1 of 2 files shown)


dkent600 (changes requested):
1) can you please supply screenshots of the changes, thanks
2) the label should not wrap unless necessary:

image

dkent600 (changes requested):
Can you please include a screenshot of what it looks like when there are a lot of social media links?

IonelLupu:
@dkent600  added another screenshot. Unfortunately, I had to add a hack to make it fit up to 9 external links. Without the hack (you will see the ternary operator in HTML) it can fit only 7 external links. Not sure why do we have to support this many in the UI, because the registration wizard has a hard condition for up to 5 external links.

dkent600:
If there if is a limit of 5 then 5 is all we need support From: Ionel lupu ***@***.***>Sent: Sunday, April 17, 2022 2:17 PMTo: PrimeDAO/prime-deals-dapp ***@***.***>Cc: Doug Kent ***@***.***>; Mention ***@***.***>Subject: Re: [PrimeDAO/prime-deals-dapp] swapped order of deal info dao name labels (PR #681) 
@dkent600  added another screenshot. Unfortunately, I had to add a hack to make it fit up to 9 external links. Without the hack (you will see the ternary operator in HTML) it can fit only 7 external links. Not sure why do we have to support this many in the UI, because the registration wizard has a hard condition for up to 5 external links.

—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>
Get Outlook for iOS

dkent600 (approved):

PR #691, opened 2022-04-18
Fix deep-diff

dkent600:
## What was done

cloned the deep-diff package into ApplyDiffService.  Fixed it to add items to arrays in a way that Aurelia can observe.

#### After


hiaux0 (approved):
Works!

applying diff now splices the array, so Aurelia can observe it!

Reason:

> How to trigger non-observable mutations such as .length or [index] =

> Every type of array mutation (including directly setting items at certain indices) can be constructed with .splice() and this is the recommended way to clear an array or to replace an item in it.

> To clear an array (as [@bigopon]() pointed out): use .splice(0). This is not a hack/workaround - this is the normal, spec-compliant (see point 6) and intended way of clearing an array.

https://github.com/aurelia/binding/issues/710#issuecomment-417794913

PR #694, opened 2022-04-19
Bugfix/claiming

vipergtsrz:
Fixed up a lot of the claiming component update issues after being able to test the contract claiming functionality


src/entities/DealTokenSwap.ts
```diff
@@ -571,9 +572,9 @@ export class DealTokenSwap implements IDeal {
   private async hydrateDaoClaims(): Promise<void> {
     const daoTokenClaims = new Map<IDAO, Array<IDaoClaim>>();
 
-    daoTokenClaims.set(this.primaryDao, await this.getDaoClaims(this.primaryDao));
+    daoTokenClaims.set(this.primaryDao, await this.getDaoClaims(this.primaryDao, this.partnerDao));
```
dkent600, line 575: would be good to comment this

src/entities/DealTokenSwap.ts
```diff
@@ -760,6 +761,7 @@ export class DealTokenSwap implements IDeal {
           if (receipt) {
             this.isExecuted = true;
             this.executedAt = new Date((await this.ethereumService.getBlock(receipt.blockNumber)).timestamp * 1000);
+            this.hydrate();
```
dkent600, line 764: What specifically needs to be hydrated here?

dkent600 (changes requested):

src/entities/DealTokenSwap.ts
vipergtsrz: Added comments

src/entities/DealTokenSwap.ts
vipergtsrz: Removed hydrate(). This was used for testing to see why another component wouldn't update and I forgot to remove it. Thanks for catching that!

dkent600 (approved):

PR #705, opened 2022-04-19
699 - fix(discussions): obs clause text on thread view

hiaux0:
## What was done
- refac to get `clauses` from the parent component.
- parent then passes into `discussionsList` and `discussionsThread` (the latter I missed in the first observability PR)

#### After

https://user-images.githubusercontent.com/30693990/164058056-f123e5c1-2290-4ef8-8fb0-5a8e0e750ca4.mp4

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
@@ -10,7 +10,7 @@
       </div>
     </section>
     <hgroup class="title" ref="refTitle">
-      <pre><h2 data-test="thread-header" class="heading topic">${dealDiscussion.topic}</h2></pre>
+      <pre><h2 data-test="thread-header" class="heading topic">${clauses.get(discussionId).text}</h2></pre>
       <span if.bind="dealDiscussion && dealDiscussion.createdBy.address">
         <h3 class="subtext">Clause #${clauseIndex}</h3>
         <h3 class="subtext">
--- a/src/dealDashboard/dealDiscussion/dealDiscussion.html
+++ b/src/dealDashboard/dealDiscussion/dealDiscussion.html
@@ -4,12 +4,14 @@
   <discussion-thread
     deal.bind="deal"
     authorized.bind="authorized"
+    clauses.bind="clauses"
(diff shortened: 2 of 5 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
```diff
@@ -10,7 +10,7 @@ <h1 class="subtitle heading heading1 gradient">
       </div>
     </section>
     <hgroup class="title" ref="refTitle">
-      <pre><h2 data-test="thread-header" class="heading topic">${dealDiscussion.topic}</h2></pre>
+      <pre><h2 data-test="thread-header" class="heading topic">${clauses.get(discussionId).text}</h2></pre>
```
dkent600, line 13: are you sure this will update the way it should?  function calls don't observe...

dkent600 (changes requested):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
hiaux0: Yup, we are getting a reference to one clause here.
That's the "trick" you suggested with the clauses getter

```
 @computedFrom("deal.registrationData.terms.clauses.length")
  private get clauses(): Map<string, IClause> {
    return new Map<string, IClause>(this.deal.registrationData.terms.clauses.map(clause => [clause.id, clause]));
  }
```

hiaux0: If adding/removing, then `...terms.clauses.length` re-triggers a new map, thus clauses.get() gets re-evaluated

hiaux0: Verified the fix:

1. Add new clause
2. --> gets updated in UI
3. Change clause text from 1.
4. --> gets updated in UI
5. Delete another clause
6. --> gets updated in UI
7. Change clause text from 1.
8.  --> gets updated in UI

dkent600: kk

dkent600 (approved):

PR #709, opened 2022-04-19
708 - Don't re-add comment at bottom

hiaux0:
## What was done
- Add `else` to catch the "original comment not deleted" case (This case happens, when eg. not signing)
- quickfix: related issue, where loading indicator was removed too quickly

## Note
This is almost impossible to reproduce in reality
Why impossible?
1. Deleting a comment gets streamed/websocketed, so we delete comment in any case
2. Happens in the short interval, where 2 persons perform delete, resp. vote, at the same time.

## Testing

1. AccountA deletes comment C
3. AccountB likes C

#### Before
4. B gets popup, BUT C gets re-added at bottom

https://user-images.githubusercontent.com/30693990/164077851-a5ea004b-0709-4d6f-9b35-2dcc7d96e452.mp4

#### After
3. B should get popup, and C gets removed immediately

https://user-images.githubusercontent.com/30693990/164077890-6006756d-4912-4d1f-ba32-5f923045cdf8.mp4

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -421,7 +421,6 @@ export class DiscussionThread {
 
     this.threadDictionary[_id] = message;
     this.updateThreadsFromDictionary();
-    this.isLoading[`isVoting ${_id}`] = false;
   }
 
   async deleteComment(_id: string): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -417,7 +421,6 @@ export class DiscussionThread {
 
     this.threadDictionary[_id] = message;
     this.updateThreadsFromDictionary();
-    this.isLoading[`isVoting ${_id}`] = false;
```
hiaux0, line 420: Ninja fix: Loading indicator removed too quickly. It did not wait for promise.

dkent600 (approved):

PR #720, opened 2022-04-20
fixed a bug that doesn't allow a deal to be saved

IonelLupu:
## What was done
fixed a bug that doesn't allow a deal to be saved if `cliffOf` and `vestedFor` were undefined
#### After

### DIFF
--- a/src/services/FirestoreService.ts
+++ b/src/services/FirestoreService.ts
@@ -99,8 +99,17 @@ export class FirestoreService<
         delete registrationData.partnerDAO;
       }
 
+      registrationData.primaryDAO.tokens.concat(registrationData.partnerDAO?.tokens ?? []).forEach(token => {
+        if (token.cliffOf === undefined) {
+          delete token.cliffOf;
+        }
+        if (token.vestedFor === undefined) {
+          delete token.vestedFor;
+        }
+      });
+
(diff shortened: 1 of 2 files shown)


src/services/FirestoreService.ts
```diff
@@ -99,8 +99,17 @@ export class FirestoreService<
         delete registrationData.partnerDAO;
       }
 
+      registrationData.primaryDAO.tokens.concat(registrationData.partnerDAO?.tokens ?? []).forEach(token => {
```
dkent600, line 102: this looks like a hack.  cliffOf and vestedFor should never be undefined, nor should they be deleted.  they should be 0.

dkent600 (changes requested):

src/services/FirestoreService.ts
IonelLupu: I thought about this but there is a small issue: by default these two fields have no value making the input not show anything when disabled, which is OK. If we set the value to be 0 on Save, that value will be saved in the DB. This means, if you go back to the edit page, you won't see empty inputs anymore but inputs with the value zero even if they are disabled.

If you are ok with that, I can se the inputs to be zero by default. 

The solution in the task is similar to what was done for the rest of the similar inputs (because we cannot save undefined in the DB)

dkent600 (approved):

Issue #721, opened 2022-04-20
Should move hacks in updateTokenSwapRegistrationData to the appropriate wizard stages


renckorzay:
![image.png](https://images.zenhubusercontent.com/609a98a27efca047604a4836/e4882465-4bd2-4c94-925c-2487c440a71a)

IonelLupu:
@paweljanicki is there a way to move these somewhere else?

I had some code here as well but I moved it in the tokenDetailsStage component. It was doing some checks and based on some condition, it deleted some fields so that Firebase won't error out.

I assume the code in the image above is for the same reason.

PR #726, opened 2022-04-20
Quick fix to avoid clause discussion with null id

paweljanicki:
#### After


dkent600 (approved):
@paweljanicki Approved for now, but will have to be refactored, see https://github.com/PrimeDAO/prime-deals-dapp/issues/721.

PR #728, opened 2022-04-20
723 - fix(terms): generate new ids for Make an offer wizard

hiaux0:
## What was done
- Add another if for "Make an offer" case when generating ids

Analysis
```
Problem: Discussions of Partnered Deal are mirrored in Open Proposal and vice-versa.
Cause: clauseId of clause in Open Proposal gets copied over to Clause of Partnered Deal

Looks like, when we overlooked this use case.
- When we create a Partnered Deal, the clauseId just get copied over.
- Which then translates to same discussionsIds
- Thus discussions are mirrored

Solution: Regenerate clauseIds when creating a Partnered Deal.
Option 1: FE creates new ids in a Partnered Deal wizard
Option 2: Firebase does this
```

## Testing
1. Make an offer to an Open Proposal
2. Verify in Firestore, that `clauseId`s are different in both deals

#### Before
Clauses have same ids.
Screenshot shows Open Proposal; Note the `clauseId`
image

#### After
Clauses have different ids
Screenshot shows Partnered Deal (created based on the above Open Proposal); Note the different `clauseId`
image

### DIFF
--- a/src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
+++ b/src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
@@ -29,8 +29,9 @@ export class TermsStage implements IBaseWizardStage {
     this.stageMetadata = stageMeta.settings;
     this.stageMetadata.termsViewModes = this.stageMetadata.termsViewModes ?? this.getDefaultTermsViewModes(stageMeta.wizardType);
 
+    this.addIdsToClauses(stageMeta.wizardType);
+
     this.wizardService.registerStageValidateFunction(this.wizardManager, async () => {
-      this.addIdsToClauses();
       this.checkedForUnsavedChanges();
       const formsAreValid = await areFormsValid(this.termClauses.map(viewModel => viewModel.form));
       return formsAreValid && !this.hasUnsavedChanges;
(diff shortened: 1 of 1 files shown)


src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
```diff
@@ -29,8 +29,9 @@ export class TermsStage implements IBaseWizardStage {
     this.stageMetadata = stageMeta.settings;
     this.stageMetadata.termsViewModes = this.stageMetadata.termsViewModes ?? this.getDefaultTermsViewModes(stageMeta.wizardType);
 
+    this.addIdsToClauses(stageMeta.wizardType);
+
     this.wizardService.registerStageValidateFunction(this.wizardManager, async () => {
-      this.addIdsToClauses();
       this.checkedForUnsavedChanges();
```
hiaux0, line 35: My fix could be debatable.
Since generating the ids is not costly, the "After" approach is "more readable" code wise.

**Before:** Only generate ids when either
- proceeding
- clicking on the step in the stepper

**After: **
- generate ids when the stage is `activated`

The other approach would have been.
1. keep `this.addIdsToClauses()` in line 33.
2. In line 32 add the if check for make-an-offer type

- This would be 2 places where `this.addIdsToClauses` would be called

hiaux0 (commented):
.

src/wizards/tokenSwapDealWizard/stages/termsStage/termsStage.ts
```diff
@@ -59,9 +59,13 @@ export class TermsStage implements IBaseWizardStage {
     this.hasUnsavedChanges = this.termClauses.filter(viewModel => viewModel.viewMode === "edit").length > 0;
   }
 
-  addIdsToClauses() {
+  addIdsToClauses(wizardType: WizardType) {
```
dkent600, line 62: functions should have return types

dkent600 (approved):

PR #755, opened 2022-04-25
arranged the UI a bit in the deal info card

IonelLupu:
## What was done
image

#### After

### DIFF
--- a/src/dealDashboard/dealInfo/dealInfo.html
+++ b/src/dealDashboard/dealInfo/dealInfo.html
@@ -14,7 +14,7 @@
           </div>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


dkent600 (changes requested):
1) Not vertically centered:

image

2) Is odd to me that this text is wrapping:

image

3) Is missing "Time Left to Initiate" (which BTW, should be "Time Left to Execute")

IonelLupu:
@dkent600 

1. fixed
2. that's the requirement. What do to?
3. There is one but it only shows up when the deal is funding.

dkent600 (approved):

PR #806, opened 2022-04-27
Update firebase authentication message

paweljanicki:
## What was done
- Update authentication message
- Don't request part of the message to sign from firebase function - create message to sign (aka authentication message) entirely on the frontend
- Send message, alongside address and signature to firebase function `verifySignedMessageAndCreateCustomToken`
- Store user signatures in local storage
- Show "check metamask" message when user is required to sign

## Testing
- connect, disconnect, switch accounts

#### Before
- signature could be used only once

#### After
- signature can be reused

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -55,9 +55,9 @@ export class App {
       this.handleScrollEvent();
     });
 
-    this.eventAggregator.subscribe("deals.loading", async (onOff: boolean) => {
-      this.modalMessage = "Thank you for your patience while we initialize for a few moments...";
-      this.handleOnOff(onOff);
+    this.eventAggregator.subscribe("deals.loading", async (config: {onOff: boolean, message?: string}) => {
+      this.modalMessage = config.message || "Thank you for your patience while we initialize for a few moments...";
+      this.handleOnOff(config.onOff);
     });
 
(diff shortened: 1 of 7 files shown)


dkent600 (changes requested):
don't show metamask message unless needed

src/app.ts
```diff
@@ -105,7 +105,12 @@ export class App {
       }
     });
 
+    this.eventAggregator.subscribe("database.account.signature.successful", () => {
+      this.modalMessage = "Thank you for your patience while we initialize for a few moments...";
```
hiaux0, line 109: Same text string is used in multiple places -> should add a constant imo.
(Future: think this message is also used in other files, so we should add a concept of "text copies" as a file or directory)

hiaux0 (commented):
## Behavior
### Local storage
> Store user signatures in local storage

1. Could you add more explanation around what the code is doing?
2. Do we only check if signature present in localStorage without another verification to firebase?
3. How do sessions work now? (ie. how long is a signature valid?)

Overall I'd like to have more documentation around the changes

### Disconnect
When I disconnect from Metamask, then reconnect, I don't get a firebase signature request.
But I should?

dkent600 (approved):

PR #841, opened 2022-04-28
811-Token Swap Status: Images won't round (deposits table)

giladt:
## What was done
Added rounding to icons at the amount column in the funding page deposits table


dkent600 (approved):

dkent600:
@giladt These repeated places where that icon is is style the same might best have a single definition of the style

PR #852, opened 2022-04-28
703 - Discussions: loading indicator when replying

hiaux0:
## What was done
- add loading indicator
  - note, it will indicate loading for _all_ comments. 
  - Imo a nice feature from network perspective to block further requests
  - though might be annoying for user if loading times are long
    - imo good balance, as we are not super confident around theconvo network yet
- and disable the reply button

## Testing
reply to a comment

#### Before
no loading indicator

#### After
image

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
@@ -107,6 +107,7 @@
           type="primary"
           full-width
           disabled.bind="isLoading.commenting || !comment.length"
+          is-loading.bind="true"
           loading.bind="isLoading.commenting"
           click.delegate="addComment()">
           ${replyToComment ? "Reply" : "Comment"}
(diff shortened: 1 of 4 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.html
```diff
@@ -107,6 +107,7 @@ <h3 class="subtext">
           type="primary"
           full-width
           disabled.bind="isLoading.commenting || !comment.length"
+          is-loading.bind="true"
```
giladt, line 110: Why is it here?
Maybe something like:
`is-loading.bind="isLoading.commenting || isLoading.replying"`?

giladt (changes requested):

giladt (changes requested):
reply

Two points to consider:
- When we press cancel- it fetches the comment again. Probably best to skip it.
- While moving the cursor away from the comment, there's no indication of loading. Not sure where will it be the best to indicate it though, op.1 next to the like icon, op2. next to the "comment" title at the bottom?

hiaux0:
> Two points to consider:
    When we press cancel- it fetches the comment again. Probably best to skip it.
    While moving the cursor away from the comment, there's no indication of loading. Not sure where will it be the best to indicate it though, op.1 next to the like icon, op2. next to the "comment" title at the bottom?

Good points, I'll create a ticket for that.

Would like to keep this PR to what was requested:
> decision: yes to disabling the button and to providing a spinny icon (in the button)

PR #855, opened 2022-04-28
811-Token Swap Status: Images won't round (All token logos in deal dashboard)

giladt:
## What was done
Added new style - `roundedLogo`
Applied this new style to all token-logos in deal dashboard:
- status card
- funding grid
- funding
- dao-icon-name

### DIFF
--- a/src/styles/styles.scss
+++ b/src/styles/styles.scss
@@ -189,6 +189,11 @@ html {
     height: 18px;
     border-radius: 50%;
   }
+
+  .roundedLogo {
+    border-radius: 50%;
+  }
+
 }
 
 @media screen and (max-width: 1072px) {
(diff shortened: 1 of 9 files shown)


src/styles/styles.scss
```diff
@@ -189,6 +189,11 @@ html {
     height: 18px;
     border-radius: 50%;
   }
+
+  .roundedLogo {
+    border-radius: 50%;
```
dkent600, line 194: should this also specify width and height?

giladt: Each image specifies its own width and height within the components, and it differs from the classes above, so I can't reuse these. I wanted to have a global class only for rounding logos.

dkent600 (approved):

PR #870, opened 2022-04-29
702- Consider to only perform UI action, when request was successful

giladt:
## What was done
Revert instant deletion.
Added spinner while deletion is processed.

#### Before
Comments have been removed instantly from the thread. If an error occurred, the message has been added again.
#### After
A spinner is shown until deletion returns success, then the message is removed from the thread.

deleting comment


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -428,25 +428,26 @@ export class DiscussionThread {
...
-        /* If deletion did not happened, restore original comments */
-        this.threadComments = swrThreadComments;
-      } else {
+      if (isDeleted) {
+        this.threadComments = this.threadComments.filter(comment => comment._id !== _id);
+        this.threadDictionary = this.arrayToDictionary(this.threadComments);
         this.updateDiscussionListStatus(new Date(), this.threadComments.length);
         this.eventAggregator.publish("handleSuccess", "Comment deleted.");
+      } else {
+        Promise.reject("Not deleted.");
```
dkent600, line 439: do you have to return this?  what does it do?

src/dealDashboard/dealDiscussion/discussionThread/singleComment/singleComment.html
```diff
@@ -95,5 +95,9 @@
       </div>
       <pre>${comment.text}</pre>
     </section>
+    <div if.bind="loading['isDeleting ' + comment._id]" class="deleteInfo">
+      <i class="spinner fa fas fa-circle-notch fa-spin"></i>
```
dkent600, line 99: I think "spinner" and "fa-spin" may be redundant

dkent600 (changes requested):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
giladt: This should trigger the `catch` in case the `discussionsService.deleteComment` has failed.

src/dealDashboard/dealDiscussion/discussionThread/singleComment/singleComment.html
giladt: `.spinner` is for styling the color. `.fa-spin` is `Font-Awesome`'s built in for spinning animation.

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
hiaux0: Just tested again, you would indeed need to return the rejection.
Also throwing an error will also trigger the `.catch()`

hiaux0: Better to throw an error object, else the error gets swallowed, because in the `.catch()` block we are using `err.message`.
You would also actually need to add logic to handle `err.message`.
Currently, we only notify user `if (comment._id)` (line 448), and not in other cases

dkent600 (approved):

hiaux0 (commented):
as discussed
- move `isCommentPresent` to service

hiaux0 (changes requested):
## Code
Muuuuuch cleaner

## Fncality

### Abort but still deleted
- :heavy_check_mark:  Tested the test flow "Test flow (@hiaux0)" 

### Response `{succuess: false}`
- not handled yet 
Reason for requesting the change

dkent600 (changes requested):
sending back for Hiau's requests

hiaux0:
There are 4 cases we care about right?:

1. delete, success: true
2. delete, but success: false
3. delete, that got aborted (comment actually deleted)
4. delete, that got aborted (really aborted, ie. comment not even deleted)

Code handles all of them, so looks good to me.

hiaux0:
@giladt 
Can you also verify the cases a last time? :point_up:

giladt:
Manually tested:
- [X] delete, success: true
- [X] delete, but success: false
- [X] delete, that got aborted (comment actually deleted)
- [X] delete, that got aborted (really aborted, ie. comment not even deleted) _pushed fix_

hiaux0 (approved):
gooooodddiee

PR #882, opened 2022-04-29
Voting: Red Circle goes to left and right while interacting with Voting buttons

vipergtsrz:
Simplified the deal votes to only use the variables it needs which fixed the UI issues in the bug

### DIFF
--- a/src/dealDashboard/dealVotes/dealRepresentativesVotes/dealRepresentativesVotes.html
+++ b/src/dealDashboard/dealVotes/dealRepresentativesVotes/dealRepresentativesVotes.html
@@ -10,11 +10,9 @@
     </div>
 
     <div class="votes-statuses-circles">
-      <template
-        repeat.for="representative of representativesSortedByVotes(deal.representatives, deal.allVotes)"
-      >
+      <template repeat.for="vote of sortedVotes">
         <div
-          class="vote-status-circle ${getVoteSlug(deal.allVotes, representative.address) ? 'vote-' + getVoteSlug(deal.allVotes, representative.address) : ''}"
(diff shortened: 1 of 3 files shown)


dkent600 (approved):

src/dealDashboard/dealVotes/dealRepresentativesVotes/dealRepresentativesVotes.html
```diff
@@ -32,19 +30,13 @@
         text="${representative.address | smallHexString}"
       >
       </etherscanlink>
-      ${getVoteSlug(deal.allVotes, representative.address) || 'hasn’t voted'}
+      ${getVoteSlug(votes, representative.address) || 'hasn’t voted'}
       <span
-        class="representative-vote ${getVoteSlug(deal.allVotes, representative.address) ? 'representative-vote-' + getVoteSlug(deal.allVotes, representative.address) : ''}"
+        class="representative-vote ${getVoteSlug(votes, representative.address) ? 'representative-vote-' + getVoteSlug(votes, representative.address) : ''}"
```
dkent600, line 35: Can you please ensure that `votes` is being properly observed when votes are changed on other browsers.  You'll need the latest code from dev for this.

dkent600 (changes requested):

src/dealDashboard/dealVotes/dealRepresentativesVotes/dealRepresentativesVotes.html
vipergtsrz: @dkent600 I just verified that the votes are being properly observed when they are changed on different browsers! Works perfectly :)

dkent600 (approved):

PR #887, opened 2022-04-29
878 - styles(dashboard.menubar): align "share" to right, and mobile fixes

hiaux0:
#### After
- All actions visible
1: before
2: after
image

- Share button (case, if only button in UI)
image

- Mobile
image

### DIFF
--- a/src/dealDashboard/dealMenubar/dealMenubar.html
+++ b/src/dealDashboard/dealMenubar/dealMenubar.html
@@ -10,11 +10,14 @@
     <i class="fas fa-share-alt"></i>
     SHARE
   </a>
-  <a
-    if.bind="canEdit"
-    route-href="route.bind: deal.isOpenProposal ? 'editOpenProposal' : 'editPartneredDeal' ; params.bind: { id: deal.id, stageRoute: 'proposal' }">
-    <pbutton type="secondary">Edit deal</pbutton>
-  </a>
 
-  <pbutton click.delegate="cancelDeal()" if.bind="canCancel" type="tertiary">Cancel deal</pbutton>
+  <div class="dealActionButtons" if.bind="canEdit || canCancel">
+    <a
+      if.bind="canEdit"
+      route-href="route.bind: deal.isOpenProposal ? 'editOpenProposal' : 'editPartneredDeal' ; params.bind: { id: deal.id, stageRoute: 'proposal' }">
+      <pbutton type="secondary">Edit deal</pbutton>
+    </a>
+
+    <pbutton click.delegate="cancelDeal()" if.bind="canCancel" type="tertiary">Cancel deal</pbutton>
+  </div>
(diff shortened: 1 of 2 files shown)


src/dealDashboard/dealMenubar/dealMenubar.html
```diff
@@ -10,11 +10,14 @@
...
     SHARE
   </a>
-  <a
-    if.bind="canEdit"
-    route-href="route.bind: deal.isOpenProposal ? 'editOpenProposal' : 'editPartneredDeal' ; params.bind: { id: deal.id, stageRoute: 'proposal' }">
-    <pbutton type="secondary">Edit deal</pbutton>
-  </a>
 
-  <pbutton click.delegate="cancelDeal()" if.bind="canCancel" type="tertiary">Cancel deal</pbutton>
+  <div class="dealActionButtons" if.bind="canEdit || canCancel">
```
hiaux0, line 14: Split up "Share" from [ Edit+Cancel ]
by wrapping Edit+Cancel with its own `div`

When thinking about the problem, this emerged as the most fitting solution.
Edit+Cancel are variable, whereas "Share" is always present, thus makes sense to split up Edit+Cancel

src/dealDashboard/dealMenubar/dealMenubar.scss
```diff
@@ -1,36 +1,48 @@
 @import "src/resources/elements/primeDesignSystem/variables";
 
 deal-menubar {
-  display: grid;
-  grid-template-columns: auto auto auto;
-  align-items: center;
+  display: flex;
```
hiaux0, line 4: flex for 
Share [Edit Cancel]

src/dealDashboard/dealMenubar/dealMenubar.scss
```diff
@@ -1,36 +1,48 @@
...
   justify-content: end;
-  grid-column-gap: 10px;
-  grid-row-gap: 20px;
 
-  i.fas {
-    font-size: 20px;
+  .dealActionButtons {
+    display: grid;
```
hiaux0, line 8: grid for 
[Edit Cancel]

Note, after this fix, might make sense to not move grid here in the first place, but I didn't want to do tooo many changes

src/dealDashboard/dealMenubar/dealMenubar.scss
```diff
@@ -1,36 +1,48 @@
...
+    grid-template-columns: auto auto;
+    align-items: center;
+    justify-content: end;
+    grid-column-gap: 10px;
+    grid-row-gap: 20px;
+
+    margin-left: 30px;
```
hiaux0, line 15: before: margin-right from the Share button.
after: margin-left from [Edit Cancel]

dkent600 (approved):

PR #896, opened 2022-05-02
moved the primary DAO logo on top of the partner DAO logo

IonelLupu:
## What was done
image

#### After


vipergtsrz:
@IonelLupu Can you please put these styles into the component so it has the same look everywhere? The component name is dao-icon-name. Changing the styles here will affect every place the site is using it not just the deal summary.

dkent600 (approved):

PR #897, opened 2022-05-02
871 - Pop-up Modals needs adjustments

hiaux0:
## What was done
- tweaked buttons styles (not up to date with design system)
- need rgba for background blur
- update dialog footer button margins

### Whitespace issue
There is a weird issue with whitespaces for `display: inline-block` (we had `display: block`, but whitespace was still there)
--> Suggestion is to use `display: flex`
https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#solutions

| Before | After |
| ------ | ----- |
| unwanted `Whitespace` node with width of 4px | no gap between elements |
| image  |  image|

## Testing
- Use chromium based browser
- Open a dialog (quickest one I found was on the Dasboard page -> Cancel Deal)

|          | Before | After |
| ------- | ------    | -----   |
| Buttons | image | image |
| Blur | image | image |

### DIFF
--- a/src/resources/dialogs/dialogs.scss
+++ b/src/resources/dialogs/dialogs.scss
@@ -6,9 +6,9 @@
  * write styles against that.  See how Alert and disclaimer dialogs are created.
  */
 body > ux-dialog-overlay.active {
-  opacity: 0.5;
-  background-color: $Border01;
+  background-color: rgba(21, 6, 45, 0.5); // rgba otherwise does not work properly with blue
   backdrop-filter: blur(10px);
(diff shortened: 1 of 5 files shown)


dkent600 (changes requested):
Awaiting confirmation that all the various popups still work

hiaux0:
- everywhere
  - tos
- wizard
  - submit
- funding
  - withdraw


- dashboard
  - make an offer
  - start funding
  image
  - delete comment
  - execute
  image

dkent600 (approved):

PR #911, opened 2022-05-03
750 - fix(discussions.thread): add new author to profile

hiaux0:
## What was done
- when new address enters the conversation, we did not update the profiles map -> update it on new comment (if author not already present)

## Testing
1. Create a _new_ discussion
2. add comment
3. reply to that comment (your own comment)

#### Before
wrong avatar

#### After
https://user-images.githubusercontent.com/30693990/166440929-bd31bec3-4268-4dbe-8ef5-6647a8543421.mp4

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -354,20 +354,21 @@ export class DiscussionThread {
   }
 
   private addAuthorToThreadProfiles(comment: IComment): void {
-    if (this.threadProfiles[comment.author]) return;
-
-    if (comment.authorENS /* author has ENS name */) {
-      this.threadProfiles[comment.author] = {
-        name: comment.authorENS,
-        address: comment.author,
-        image: "",
-      };
-    } else {
-      /* required for replies */
-      this.discussionsService.loadProfile(comment.author).then(profile => {
-        this.threadProfiles[comment.author] = profile;
-        this.isLoading[comment.author] = false;
-      });
+    if (!this.threadProfiles[comment.author]) {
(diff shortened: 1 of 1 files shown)


giladt (changes requested):

src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -352,6 +338,25 @@ export class DiscussionThread {
...
+        this.threadProfiles[comment.author] = {
+          name: comment.authorENS,
+          address: comment.author,
+          image: "",
+        };
+      } else {
+        /* required for replies */
+        this.discussionsService.loadProfile(comment.author).then(profile => {
+          this.threadProfiles[comment.author] = profile;
+          this.isLoading[comment.author] = false;
```
hiaux0, line 354: Wondering why it's enough to have this change to `= false` in the `else` statement only?

loading `= true` is set before the `if` in line 343.
So I expect `= false` at the end of the `if`, ie. after line 356

@giladt

giladt: We want to set the loading to false only after the `loadProfile` promise is fulfilled. Otherwise, it will be set to `false` immediately.

But, you are right- 
- Setting `isLoading[comment.author] = true` can be moved into the `else` statement, since it's the only async operation, and 
- the setting to `false` should probably be better inside a `.finally()` statement... 
- (Applies also for line 240...)
@hiaux0

hiaux0: Looks like a minor improvement then.
Will skip

giladt (approved):

PR #912, opened 2022-05-03
784- Replies Column display the total amount of comments happened while deal was private.

giladt:
## What was done
Split replies count in `clauseDiscussion` into:
 - replies: total count of public and private comments
 - publicReplies: count of only public comments (metadata.isPrivate === "false")

The discussions list shows the replies count based on the deal authentication state:
If a user is authenticated as a representative/proposal lead- will show the `replies` count, otherwise the `publicReplies`.

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -185,7 +185,7 @@ export class DiscussionThread {
       };
     }
     this.updateThreadsFromDictionary();
-    this.updateDiscussionListStatus(new Date(), this.threadComments?.length || 0);
+    this.updateDiscussionListStatus(new Date());
 
     // scroll to bottom only if the user is at seeing the last message
     if (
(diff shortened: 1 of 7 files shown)


hiaux0 (changes requested):
## code
- looks good overall
- missed one place: 
`if.to-view="!discussionsService.discussions[clause.id] || discussionsService.discussions[clause.id].replies === 0">Begin A Discussion</div>`
Would it be enough to use ?:
```diff
- discussions[clause.id].replies
+ discussions[clause.id].publicReplies
```

## Fnaclity
- generally works
- works when disconnecting
- :warning: not working for changing accounts (or reconnecting to address, that can see private comment)


video

dkent600 (approved):

giladt:
> missed one place in dealClauses.html

- If it's an open deal- all threads are public, so 'replies' and 'publicReplies' should always be the same since the deal can not be private.
- If it's a partnered deal- only representatives/proposal lead can discuss, thus by definition see private discussions (if connected). so 'replies' is correct and shows the total count of public+private comments.
- If no user is not connected, then the `discuss`/`begin a discussion` button is not visible.

giladt:
> * ⚠️ not working for changing accounts (or reconnecting to address, that can see private comment)

Will be handled in a separate issue / PR: #918

hiaux0 (approved):
> If it's a partnered deal- only representatives/proposal lead can discuss, thus by definition see private discussions (if connected). so 'replies' is correct and shows the total count of public+private comments.

Makes sense :heavy_check_mark:

PR #920, opened 2022-05-04
Fix errors on primary DAO wizard stage, remove hack for missing partn…

paweljanicki:
…erDAO, and improve the hack for missing funding period

#### After


src/wizards/tokenSwapDealWizard/stages/primaryDaoStage/primaryDaoStage.ts
```diff
@@ -22,7 +22,9 @@ export class PrimaryDaoStage implements IBaseWizardStage {
     this.disabled = stageMeta.wizardType === WizardType.makeAnOffer;
     this.isPartneredDeal = this.getIsPartneredDeal(stageMeta.wizardType);
 
-    const validationRules = daoStageValidationRules("Primary DAO", this.wizardState.registrationData.partnerDAO);
+    const partnerDao = this.isPartneredDeal ? this.wizardState.registrationData.partnerDAO : null;
```
hiaux0, line 25:
```suggestion
    const partnerDao = this.isPartneredDeal ? this.wizardState.registrationData.partnerDAO : undefined;
```

Because the type `IDealRegistrationTokenSwap` does not specify `null`
`partnerDAO?: IDAO; // This doesn't exist for an Open Proposal`

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
```diff
@@ -80,14 +80,21 @@ export class TokenDetailsStage {
...
-          !this.hasUnsavedChangesForPrimaryDetails &&
-          !this.hasUnsavedChangesForPartnerDetails &&
-          this.hasValidPartnerDAOTokensDetailsCount &&
-          primaryTokensValid &&
-          (this.isOpenProposalWizard ? true : partnerTokensValid),
-        );
+        .then(async (result) => {
+          if (result.valid) {
+            if (this.wizardState.registrationData.fundingPeriod === undefined) {
+              this.wizardState.registrationData.fundingPeriod = null;
```
hiaux0, line 86: Before: `delete registrationData.fundingPeriod;`
After: `.fundingPeriod = null`

Can you elaborate?

src/wizards/tokenSwapDealWizard/stages/submitStage/submitStage.html
```diff
@@ -111,7 +111,7 @@ <h4 class="submitCardHeader">Primary DAO</h4>
           </section>
 
           <section class="stageSectionContent">
-            <pcard if.bind="submitData.partnerDAO">
+            <pcard if.bind="!isOpenProposalLike">
```
hiaux0, line 114: Why this change?
It makes sense, but I would prefer the previous one, as the section is about "Partner DAO", and the `if.bind="submitData.partnerDAO"` is a bit more descriptive

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
paweljanicki: Yes, we set it to `null` in order to fix a bug when resetting fundingPeriod. It was impossible to reset funding period after it was set (for open proposal), for example trying to change from 100 day to nothing (empty input). FundingPeriod in Firestore was remaining untouched. That's why now we save it as `null`

src/wizards/tokenSwapDealWizard/stages/submitStage/submitStage.html
paweljanicki: If we use `submitData.partnerDAO` here in the template it adds `partnerDAO` property to `submitData` with value `undefined`. We don't want to have `partnerDAO` property at all

dkent600: I really doubt that just referencing a property in if.bind would cause the property to be altered.

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.ts
dkent600: You've confirmed that undefined and null behave the same everywhere in the app wrt partnerDAO?

dkent600 (changes requested):

src/wizards/tokenSwapDealWizard/stages/primaryDaoStage/primaryDaoStage.ts
```diff
@@ -22,7 +22,9 @@ export class PrimaryDaoStage implements IBaseWizardStage {
     this.disabled = stageMeta.wizardType === WizardType.makeAnOffer;
     this.isPartneredDeal = this.getIsPartneredDeal(stageMeta.wizardType);
 
-    const validationRules = daoStageValidationRules("Primary DAO", this.wizardState.registrationData.partnerDAO);
+    const partnerDao = this.isPartneredDeal ? this.wizardState.registrationData.partnerDAO : undefined;
```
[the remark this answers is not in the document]
hiaux0: It was actually `null`, but I asked for a change https://github.com/PrimeDAO/prime-deals-dapp/pull/920#discussion_r864607488

For `null` we would ideally also update the type in `IDealRegistrationTokenSwap`
```ts
partnerDAO?: IDAO; // This doesn't exist for an Open Proposal
```
that's why I preferred `undefined`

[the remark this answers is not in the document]
hiaux0: Wait, I had a thinking mistake from the start.
We should rather update the following and not `IDealRegistrationTokenSwap`
```diff
- export const daoStageValidationRules = (title: string, otherDao?: IDAO) =>
+ export const daoStageValidationRules = (title: string, otherDao: IDAO | null) =>
```

src/wizards/tokenSwapDealWizard/stages/submitStage/submitStage.html
paweljanicki: I was shocked but it did create the property. According to my debugging, referencing a non-existing property in if.bind, adds that property to the object with value undefined

dkent600: Maybe that is so it has something to observe.  I never realized that was how it did it.

hiaux0: I can see this being a potenial bug, but I couldn't reproduce this "adds non-existing prop" behavior

STEPS
1. readd if.bind="submitData.partnerDAO"
2. in bind() hook add `console.log(JSON.stringify(submitData, null, 4));
3. in attached() hook add the same `console.log(JSON.stringify(submitData, null, 4));
RESULT
4. no `partnerDAO` prop was added

dkent600 (approved):

PR #928, opened 2022-05-04
918-Discussions list should update replies count when re connecting to a wallet

giladt:
## What was done
1. Simplify the filter to use deal.isAuthenticated…
2. Add a ‘snapshot’ flag to return an unfiltered comment array (and also not decrypted) - used to update the correct count. 
3. Updating the discussion list to use ‘authorized’ bindable.
4. Some cleanup


dkent600:
Deferring to V2

PR #929, opened 2022-05-04
892- Discussions: Improve loading indicator for single comments

giladt:
## What was done
- [X] Spinner is showing also while not hovering over a comment
- [X] Spinner shows only for the relevant comment.

### Before
Spinner was shown only while hovering a comment (all comments in the thread)

### After
reply loading spinner


giladt:
Should I refactor the `voting` (aka Like/Dislike) spinner to be positioned in the same way, to maintain consistency?
@dkent600 @hiaux0

dkent600 (approved):

PR #935, opened 2022-05-04
933-Discussions: Improve loading indicator for single comments - Like/Dislike

giladt:
## What was done
Fix comment 'like/dislike' button spinner to be consistent with the 'reply' button spinner and show both on hover and while the mouse is away.

like-dislike-spinner


hiaux0 (approved):
good idea with moving spinner outside!

voting triggers spinner :heavy_check_mark:

dkent600 (approved):

PR #944, opened 2022-05-05
926- Don't fetch comment when cancelling a reply

giladt:
## What was done
check if a comment exists only if initiating a reply.

### DIFF
--- a/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
+++ b/src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
@@ -355,31 +355,31 @@ export class DiscussionThread {
   }
 
   async replyComment(_id: string): Promise<void> {
-    this.isLoading.replying[_id] = true;
-    const comment = await this.discussionsService.getSingleComment(_id);
-    this.isLoading.replying[_id] = false;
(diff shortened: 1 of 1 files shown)


src/dealDashboard/dealDiscussion/discussionThread/discussionThread.ts
```diff
@@ -355,31 +355,31 @@ export class DiscussionThread {
...
-     *
-     * 2. DOMException: theconvo sdk throws this error.
-     *   This happens, when there was no response from their endpoint within a timeout limit.
-     *   Because, there was no response, we cannot assume what happened, so just early return with error popup.
-     */
-    if ((comment as any).error instanceof DOMException) {
-      this.eventAggregator.publish("handleFailure", "An error occurred. Cannot reply to comment. Please try again later");
-      return;
-    }
+    if (!this.replyToComment) {
```
hiaux0, line 358: This was just a one line fix, but caused 23 diff changes.
Would appreciate a small comment for the reviewer, so they don't need to go through all the 23 lines.
(could be solved by better diff algos, but github does not have them yet :(

hiaux0 (approved):

PR #977, opened 2022-05-10
954 - Add tests coverage

hiaux0:
## What was done
- update coverage package (`jsdevtools/coverage-istanbul-loader` is more up to date, at least it works with webpack5)
- ignore /coverage dir
- 5k diff in `package-lock.json`, because coverage reporter uses `babel` heavily
- integrate coverage lib in cypress code

:warning: There is a bug when parsing code, that has coverage-helper-code.
More specifically, aurelia-validation does not parse the helper code correctly https://github.com/aurelia/testing/issues/94.
I found a fix, but would need to PR to the aurelia repo and stuff.
(Maybe we just fork ourselves, will create a ticket)
image

## Testing
(note, the section below is from the updated README
Have to run a separate command to instrument the code (instrument = add helper code, that tracks when code was executed)
`npm run start-dev:cov`

Then can run other e2e commands

#### After
image


cypress/integration/common/deal-api.ts
```diff
@@ -26,6 +26,11 @@ export class E2eDealsApi {
     return Cypress.firestoreService;
   }
 
+  private static getDealService() {
+    // @ts-ignore - Hack to access dealService inside Cypress
+    return Cypress.dealService;
+  }
```
hiaux0, line 32: @dkent600 
okay to have this changes in this PR, or do you prefer separate?

dkent600 (approved):

PR #982, opened 2022-05-11
973-Anchor links are scrolling behind the fixed header

giladt:
## What was done
Add `scroll-margin-top` to header elements to correctly position the page after scrolling to an anchor link.

*NOTE:* Images below show test content and not the actual documentation files!

#### Before
doc-scroll-before

#### After

doc-scroll-after

doc-scroll-mobile


ogsamovar:
Thanks!

dkent600 (approved):

PR #1012, opened 2022-05-17
648- Updated the prime deals video on the home page

giladt:
## What was done
Updated video URL

### DIFF
--- a/src/home/home.html
+++ b/src/home/home.html
@@ -77,7 +77,7 @@
         </div>
 
         <div class="heroColumn">
-          <iframe class="introVideo" src="https://www.youtube.com/embed/NJBQj_cZlP8" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+          <iframe class="introVideo" src="https://www.youtube.com/embed/FhE4kWEiCJM" title="Prime Deals - Web3 partnerships made easy" frameborder="0" allowfullscreen></iframe>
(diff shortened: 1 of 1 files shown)


src/home/home.html
```diff
@@ -77,7 +77,7 @@
         </div>
 
         <div class="heroColumn">
-          <iframe class="introVideo" src="https://www.youtube.com/embed/NJBQj_cZlP8" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+          <iframe class="introVideo" src="https://www.youtube.com/embed/FhE4kWEiCJM" title="Prime Deals - Web3 partnerships made easy" frameborder="0" allowfullscreen></iframe>
```
hiaux0, line 80: Why were these attributes removed?
`allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture`

hiaux0 (changes requested):

src/home/home.html
giladt: because all of them seem to
1. be irrelevant.
2. not having any effect on the outcome. (Or maybe on mobile? I didn't check).

hiaux0 (approved):
image

src/home/home.html
hiaux0: > I don’t know where those args came from, but I didn’t see that in google documentation…
https://developers.google.com/youtube/iframe_api_reference

PR #1019, opened 2022-05-20
AU2: Updated webpack to serve static in build

giladt:
## What was done
Added `copy-webpack-plugin` package
Added `serve` package
Added scripts `build`, `build-dev`, `serve` and `serve-dev`

## Why
`build`- Builds the application in production mode
`build-dev`- Builds the application in development mode
`serve`- Serves the build on localhost (to test if the build runs correctly)
`serve-dev`- Serves the app in development mode. Similar to `start`.


hiaux0 (commented):
@giladt 
Looks good

But can you add a small description of why this needed to be done?
I can see an explanation be helpful for some future version of me

PR #1039, opened 2022-05-26
changed the text copy for unauthenticated users on the deal voting section

IonelLupu:
## What was done
Before:
image

After:
image

#### After

### DIFF
--- a/src/dealDashboard/dealVotes/dealVotes.ts
+++ b/src/dealDashboard/dealVotes/dealVotes.ts
@@ -48,23 +48,18 @@ export class DealVotes {
       statusText: "Token Swap Failed",
     },
     {
-      condition: () => !this.deal.majorityHasVoted && !this.deal.isAuthenticatedRepresentativeUser && !this.deal.isAuthenticatedProposalLead,
-      voteText: "Waiting for the representatives to vote",
-      statusText: "Voting is progress",
-    },
-    {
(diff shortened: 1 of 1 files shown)


src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -47,24 +47,19 @@ export class DealVotes {
...
-      voteText: "The deal will be approved once the majority has voted in favor. Once the deal is approved, the Proposal Lead will initiate the funding phase",
+      condition: () => !this.deal.majorityHasVoted && this.deal.isAuthenticatedProposalLead,
+      voteText: "The deal will be approved once the majority has voted in favor. Once the deal is approved, you can initiate the funding phase",
       statusText: "Voting in progress",
     },
     {
-      condition: () => !this.deal.majorityHasVoted && this.deal.isAuthenticatedProposalLead,
-      voteText: "The deal will be approved once the majority has voted in favor. Once the deal is approved, you can initiate the funding phase",
+      condition: () => !this.deal.majorityHasVoted,
+      voteText: "The deal will be approved once the majority has voted in favor. Once the deal is approved, the Proposal Lead will initiate the funding phase",
```
hiaux0, line 62: nitpick:

If you would have changed the order of the right side, it would have only showed as one diff

image

hiaux0 (approved):
image

PR #1055, opened 2022-06-09
deals-page-mobile

giladt:
## What was done
* Carousel set to horizontal: 
Logically it makes no sense to show all the 'featured cards' vertically, and below that all the deals again.
* Rearranged the table to fit into mobile screens.

As a result- there is no h-scrollbar anymore on this page.

#### Before
image

#### After
image


hiaux0 (changes requested):
If I read the PR correctly, that's a bigger style change compared to Deals v1?

Can you think about a way we can keep track of this change QA wise?

hiaux0 (approved):

giladt:
Alternate PR created for v1.0

PR #1069, opened 2022-06-21
add mobile responsiveness to the all-deals page.

giladt:
## What was done
Rearranged the table to fit into mobile screens.
Small mobile (phones) should not show the titles and the age in the deals table (reduced clutter).
As a result- there is no h-scrollbar anymore on this page.

Approved by @v0ort 

#### Before
before view

#### After
Phone:
after view- phone
Tablet:
after view- tablet

### DIFF
--- a/src/deals/list/deals.scss
+++ b/src/deals/list/deals.scss
@@ -339,6 +339,10 @@ pbutton button.secondary {
       }
 
       &.all-deals {
+        .heading02Gradient {
+          padding-bottom: $spacing-small;
+          border-bottom: 1px solid $Border01;
+        }
         .grid.dealGrid {
           grid-template-columns: 1fr;
           .row {
@@ -356,10 +360,6 @@ pbutton button.secondary {
             }
             &.header {
               display: none;
-            }
-            &:first-of-type {
-              margin-top: $spacing-small;
-              border-top: 1px solid $Border01;
             }
             .cell {
               padding: 0 $spacing-normal;
--- a/src/deals/list/deals.scss
+++ b/src/deals/list/deals.scss
@@ -62,7 +62,7 @@ pbutton button.secondary {
       text-align: right;
       margin-bottom: 28px;
       display: grid;
-      grid-template-columns: 1fr auto;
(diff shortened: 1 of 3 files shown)


src/deals/list/deals.scss
```diff
@@ -322,9 +322,82 @@ pbutton button.secondary {
   }
 }
 
+@mixin prefix($text) {
```
hiaux0, line 325: Let's keep an eye on this to be moved to a more common place

src/deals/list/deals.scss
```diff
@@ -340,3 +413,32 @@ pbutton button.secondary {
     }
   }
 }
+
+@media screen and (max-width: 400px) {
```
hiaux0, line 417: I remember we agreed on 
`$MobileWidth: 375px;`

(you can search for "375" and then in the ux/ui channel to see the conversation)

Feel free to bring up suggestions

src/deals/list/deals.scss
```diff
@@ -340,3 +413,32 @@ pbutton button.secondary {
...
+      &.all-deals {
+        .grid.dealGrid {
+          .row {
+            grid-template-areas:
+              "daos daos daos"
+              "type type type"
+              "size size status";
+            grid-template-columns: 2fr 1fr 0.5fr;
+            .title, .age {
+              display: none;
```
hiaux0, line 429: The reason for not showing title is because of the space, but imo, is Title not one of the most important piece of information for a Deal?

Though looking at the reality, the titles just reflect the DAO names.
image

If you agree on including titles in the mobile view for tables, could you create a separate ticket?
Regardless, I'm fine with skipping it now.

hiaux0 (commented):
### Fncality :heavy_check_mark: 

### Code
some suggestions and questions

src/deals/list/deals.scss
giladt: Issue here is for the range from 375px to ~550px, where the 'title' starts to break- then it might look off.
Possible solution is to change the media query to reach max 550px.
In anycase I aggree that it should be either 375px or ~550px and not 400px.

PR #1115, opened 2022-07-05
Update Staging - v1.0.6

hiaux0:
## What was done
- update firebase token flow
- fix deals table

#### After


hiaux0:
blocked by 
https://github.com/firebase/firebase-tools/issues/4698

might need to downgrade firebase-tools

> "firebase deploy --only firestore:rules" works with firebase-tools version 11.1.0, but not with version 11.2.0. Additional info...

PR #1119, opened 2022-07-05
1052 - Update gnosis auth (mainnet)

tamagosante:
## What was done
- Remove auth flow (save to do because, gnosis prevents creating tx, when user is not owner)
- Add handling of wrong network (like production app)
  - Note, when switching networks, Safe App autoconnects
- Error handling, when user is not owner
image
- Text copy for "Connect to a Wallet" button -> "Connect to Safe"
image
image

## Testing
https://prime-deals-dapp-git-feat-1052-update-gnosis-auth-curvelabs.vercel.app/

```mermaid
stateDiagram
[*] --> Initial_connect
Initial_connect --> Autoconnect_with_safe_address : (and disclaimer accepted)
Initial_connect --> Change_network_to_mainnet
    Change_network_to_mainnet --> Deals_modal_switch_networks_?
      Deals_modal_switch_networks_? --> Not_connected_1 : "NOT NOW"
          Not_connected_1 --> Error_Please_connect_to_mainnet
            Error_Please_connect_to_mainnet --> [*]
      Deals_modal_switch_networks_? --> Metamask_switch_network_? : "YES, PLEASE"
          Metamask_switch_network_? --> Not_connected_1 : "Cancel"
          Metamask_switch_network_? --> Autoconnect_with_safe_address : "Switch network"

[*] --> Re_open_deals_safe_app
  Re_open_deals_safe_app --> Already_connected_?
  Already_connected_? --> Correct_network_? : Yes
  Already_connected_? --> Correct_network_? : No
    Correct_network_? --> Deals_modal_switch_networks_? : No
    Correct_network_? --> Autoconnect_with_safe_address : Yes (ignore disclaimer flow)
      Autoconnect_with_safe_address --> [*]

[*] --> Disconnect_account
  Disconnect_account --> Not_connected_2
    Not_connected_2 --> [*]
```

#### Before
Needed to auth with Firebase

#### After
Don't need to auth with Firebase

### DIFF
--- a/src/services/EthereumService.ts
+++ b/src/services/EthereumService.ts
@@ -342,8 +342,12 @@ export class EthereumService {
 
   public async ensureMetaMaskWalletProvider(): Promise<void> {
     if (!this.metaMaskWalletProvider) {
-      const provider = detectEthereumProvider ? (await detectEthereumProvider({ mustBeMetaMask: true })) as any : undefined;
-      this.metaMaskWalletProvider = provider;
+      try {
+        const provider = detectEthereumProvider ? (await detectEthereumProvider({ mustBeMetaMask: true })) as any : undefined;
+        this.metaMaskWalletProvider = provider;
+      } catch (error) {
+        this.consoleLogService.logObject(error.message, error, "error");
+      }
     }
   }
 
@@ -403,9 +407,15 @@ export class EthereumService {
   }
 
   public async getNetworkName(provider: any): Promise<string> {
-    const chainName = this.chainNameById.get(Number(await provider.request({ method: "eth_chainId" })));
+    try {
+      const chainName = this.chainNameById.get(Number(await provider.request({ method: "eth_chainId" })));
 
-    return chainName;
+      return chainName;
+    } catch (error) {
+      this.consoleLogService.logObject(error.message, error, "error");
+      return "";
+    }
+
   }
 
   public async isWrongNetwork(): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


src/services/EthereumService.ts
```diff
@@ -349,6 +349,7 @@ export class EthereumService {
 
   private async addWalletProviderListeners(): Promise<void> {
     await this.ensureMetaMaskWalletProvider();
+    await this.removeWalletProviderListeners();
```
IonelLupu, line 352: it seems the `this.removeWalletProviderListeners();` method already has `this.ensureMetaMaskWalletProvider()` inside it. Should we just remove the one above?

hiaux0: I thought so too, but I thought, that both the `add...` and the `remove...` methods could also be called separately, so I just wanted to be overly protective.
(I was kind of experimenting with this `ensureXYZ` pattern, and I like it :D )

Maybe changing the order of the methods would make more sense?

IonelLupu: yep. changing order I think will make more sense

src/app.ts
```diff
@@ -85,8 +85,21 @@ export class App {
     this.eventAggregator.subscribe("Network.wrongNetwork", async (info: { provider: any, connectedTo: string, need: string }) => {
 
       let notChanged = true;
+
+      let message = `<p>Your wallet is connected to ${info.connectedTo ?? "an unknown network"}, but to interact with deals we need you to connect to ${info.need}.  Do you want to switch your connection ${info.need} now?<p>`;
+      let header: string | undefined = undefined;
+      if (await this.ethereumService.isSafeApp()) {
```
testing-life, line 91: can this return an error? if so, do we need to handle it somehow ?

src/resources/elements/ConnectButton/ConnectButton.ts
```diff
@@ -6,11 +6,15 @@ import { autoinject, containerless, customElement, singleton } from "aurelia-fra
 
 import { DisposableCollection } from "services/DisposableCollection";
 import { EventAggregator } from "aurelia-event-aggregator";
-import { EventConfigTransaction } from "services/GeneralEvents";
+import { EventConfigException, EventConfigTransaction } from "services/GeneralEvents";
 import { TransactionReceipt } from "services/TransactionsService";
 import { Utils } from "services/utils";
 import { bindable } from "aurelia-typed-observable-plugin";
 
+const SAFE_APP_ERROR_CODE = 200;
```
testing-life, line 14: dont suppose we handle all such consts in one place, should this be in need of changing ?

src/app.ts
hiaux0: good point, I added try/catch es to all relevant methods

src/resources/elements/ConnectButton/ConnectButton.ts
hiaux0: ideally yes, but since we don't right now, imo, the next best place is on top of the file.
I'll add a todo comment

src/app.ts
hiaux0: [0a983c3](https://github.com/PrimeDAO/prime-deals-dapp/pull/1119/commits/0a983c3b6cf95ecf1e0adb1bdf94653fb385190f)

PR #1122, opened 2022-07-06
1094 - added the "make offers private" toggle in the submit stage

IonelLupu:
## What was done
The issue was already in AU1. It's not an AU2 issue. We had to have two toggle buttons: one for open proposals and one for partnered deal. I added them both now.
#### After


hiaux0 (approved):
both toggles are now connected (lead details <> submit)

PR #1134, opened 2022-07-22
Feature/1061 clause enhancements

testing-life:
## What was done
image
image

#### After


hiaux0:
## Fnality

Test text:

**bold**, *italic*, underline

- unoreder list
    - sub
1. orede
    1. sub


### copy paste
- from notion misses underline (note, pasting the notion text into github/gdoc also misses underline, so we might be okay?)

image
image

- from gdoc --> not working

- [ ] Find out if copy paste from other sources should be supported (because "people likely write up the terms somewhere else and just copy into the wizard cc @blink-km)
  - [ ] which ones?

hiaux0:
Could you add into the initial comment 
1. why ckeditor was chosen?
2. what format the formatted text is saved with?

blink-km:
copy paste from other sources should be supported

src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
```diff
@@ -1,4 +1,4 @@
-@import "../variables";
+@import '../variables';
```
giladt, line 1: Our convention is using double quotes...

src/styles/styles.scss
```diff
@@ -1,12 +1,12 @@
-@import "./animations";
-@import "./reset";
-@import "../resources/elements/primeDesignSystem/styles";
-@import "../resources/elements/primeDesignSystem/variables";
-@import "../resources/elements/primeDesignSystem/colors";
-@import "./colors";
-@import "./variables";
-@import "tippy.js/dist/tippy.css"; // TODO: use CDN?
-@import "the-new-css-reset/css/reset.css";
+@import './animations';
```
giladt, line 1: Same, please keep the double qoutes.

src/resources/elements/primeDesignSystem/pform-input/pform-input.html
```diff
@@ -2,14 +2,10 @@
...
-    <pquestion-mark
-      class="labelInfoIcon"
-      if.bind="labelInfo"
-      text.bind="labelInfo"
-    ></pquestion-mark>
+    <pquestion-mark class="labelInfoIcon" if.bind="labelInfo" text.bind="labelInfo"></pquestion-mark>
   </span>
   <div class="formInputLabelExtra" if.bind="showCounter">
-    ${inputReference.value.length || 0}${maxLength ? '/' + maxLength : ''}
+    ${charValue || inputReference.value.length || 0}${maxLength ? '/' + maxLength : ''}
```
giladt, line 8: Please note that you're changing a design-system element here. Though integrated with this project, it might be used as a reference in other projects. Is this change applied to the DS or is it specific to the clause text area? If the latter- it may be best to implement it outside of the DS and make sure it's not implicating any of the currently used elements...

@hiaux0 - what's your opinion here?

src/resources/elements/primeDesignSystem/pform-input/pform-input.html
```diff
@@ -18,7 +14,7 @@
 <p if.bind="labelDescription" class="labelDescription">${labelDescription}</p>
 
 <!-- Input section -->
-<div class="inputContainer">
+<div class="inputContainer ${isVertical ? 'isVertical':''}">
```
giladt, line 17: same as above

giladt (changes requested):
Additionally:
### Styling:
The blue frame for focused inputs does not apply to our design system. It should be  #a258a7 ($Secondary02). Preferably it should be for the whole component (including the menubar):


Should be:


The same goes for the `link` box:


The border and buttons should match the DS. Currently, we do not have matching icons, but probably colors should do the work. (@v0ort )

src/resources/elements/primeDesignSystem/pform-input/pform-input.html
testing-life: my idea was to extend, which means other instances shouldn't be affected. if you don't pass `charValue` or  `isVertical`, then nothing happens. but if you are unconvinced, by implementing it outside of DS, you mean just making a separate component specifically for the clause fnality ?

src/resources/elements/primeDesignSystem/pform-input/pform-input.scss
testing-life: sorry, automatic formatting

testing-life:
changes addressed:

- link formatting focus and icons styling
- paste in from gdocs
- initial input height
- input resizing

testing-life:
>    why ckeditor was chosen?
documentation, freshness of packages, ease of styling, available plugins. oh and i got it working the easiest :P

>   what format the formatted text is saved with?
it's a string containing the text and formatted words wrapped in HTML elems (i,u,b, ul>li, etc)

src/wizards/tokenSwapDealWizard/stages/termsStage/termClause/custom.css
```diff
@@ -0,0 +1,118 @@
...
+  /* Helper variables to avoid duplication in the colors. */
+  --ck-custom-background: hsl(270, 1%, 29%);
+  --ck-custom-foreground: hsl(255, 3%, 18%);
+  --ck-custom-border: hsl(300, 1%, 22%);
+  --ck-custom-white: hsl(0, 0%, 100%);
+
+  /* -- Overrides generic colors. ------------------------------------------------------------- */
+
+  --ck-color-base-foreground: var(--ck-custom-background);
+  --ck-color-focus-border: #a258a7;
```
giladt, line 17: please use DS variables ($Secondary02)
You can find it under
/src/resources/elements/primeDesignSystem (/colors.scss)

src/wizards/tokenSwapDealWizard/stages/termsStage/termClause/custom.css
```diff
@@ -0,0 +1,118 @@
...
+  --ck-focus-outer-shadow: transparent;
+  /* -- Overrides the default .ck-button class colors. ---------------------------------------- */
+
+  --ck-color-button-default-background: var(--ck-custom-background);
+  --ck-color-button-default-hover-background: hsl(270, 1%, 22%);
+  --ck-color-button-default-active-background: hsl(270, 2%, 20%);
+  --ck-color-button-default-active-shadow: hsl(270, 2%, 23%);
+  --ck-color-button-default-disabled-background: var(--ck-custom-background);
+
+  --ck-color-button-on-background: var(--ck-color-button-active-bg);
```
giladt, line 38: The same color goes here ($Secondary02)
Text color should be $BG04

src/wizards/tokenSwapDealWizard/stages/termsStage/termClause/custom.css
```diff
@@ -0,0 +1,118 @@
...
+
+  /* -- Overrides the default .ck-balloon-panel class colors. --------------------------------- */
+
+  --ck-color-panel-background: var(--ck-custom-background);
+  --ck-color-panel-border: var(--ck-custom-border);
+
+  /* -- Overrides the default .ck-toolbar class colors. --------------------------------------- */
+
+  --ck-color-toolbar-background: var(--ck-custom-background);
+  --ck-color-toolbar-border: var(--ck-custom-border);
```
giladt, line 93: border-color should be $Border01, 
background-color should be $BG04.

giladt (changes requested):
1. Probably best if you can use the palette colors from the design system for all elements rather than fixed `HSL`, `hex` etc. Please note that there are also variables for sizes, such as border-radius and padding that you can use.

2. I get an error after trying to open a link I've entered. Maybe you can find a fix for that?

3. Please make the hover-border change through the whole element and not only the lower half. If it requires a longer time to refactor, maybe it's better to triage it because it would probably come back to us during testing.

4. Please revert the quotes in all files to double quotes... It appears to apply all `.scss` files.

5. Please don't hate me for being picky! 😂

testing-life:
> 1. Probably best if you can use the palette colors from the design system for all elements rather than fixed `HSL`, `hex` etc. Please note that there are also variables for sizes, such as border-radius and padding that you can use.
> 
>     2. I get an error after trying to open a link I've entered. Maybe you can find a fix for that?
>
>     3. Please make the hover-border change through the whole element and not only the lower half. If it requires a longer time to refactor, maybe it's better to triage it because it would probably come back to us during testing.
> 
>     4. Please revert the quotes in all files to double quotes... It appears to apply all `.scss` files.
> 
>     5. Please don't hate me for being picky! joy

ad1. fixed

ad2.
resolved
image

ad3. fixed
image

ad4.fixed

ad5. that's ok, i am capable of hating people for several things simultaneously :P

giladt (changes requested):
Found three more fun tasks for you (While adding a second clause):
1. Styling is gone here:



3. IMO we can remove unneeded `/build/translations` files from the repo unless there's a good reason for leaving them there.

4. This:
Please test also deleting and editing saved clauses, to see if it plays right...



The rest seems fine with me. Noice work!

testing-life:
ad1.  reinstated custom css. overriding some custom props from the component turned out impossible, plus i do think we should look into custom props grolbally, as per the telegram chat
ad2. removed
ad3. `querySelector` now uses index, so displays multiple editors.

giladt (approved):

PR #1139, opened 2022-07-28
986 - Showing total price value at execution

IonelLupu:
## What was done
After a deal is executed, it will show the price at the time of execution instead of showing the price at the current moment of page visualization:
image

Business requirements: https://www.notion.so/primedao/Prime-Deals-V1-1-Feature-set-A-85599522f25f4b75b895c466f5ad9fd4#5f0d85a7312d4ca097fcdf1b251fd1f0
## Testing
execute a deal and it should display the price at the moment of execution. Testing if this is correct implies a price change of the token which may take weeks (unless we can manually change the current token price). After the prices changes, we should see the same "total funds value" (because it was saved in our DB)

### DIFF
--- a/src/services/register.ts
+++ b/src/services/register.ts
@@ -57,7 +57,6 @@ export const register: IRegistry = {
     container.register(Registration.singleton(DisclaimerService, DisclaimerService));
     container.register(Registration.singleton(FirebaseService, FirebaseService));
     container.register(Registration.singleton(FirestoreService, FirestoreService));
-    container.register(Registration.singleton(FirestoreDealsService, FirestoreDealsService));
     container.register(Registration.singleton(EnsService, EnsService));
     container.register(Registration.singleton(AxiosService, AxiosService));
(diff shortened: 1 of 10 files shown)


src/services/register.ts
```diff
@@ -57,7 +57,6 @@ export const register: IRegistry = {
     container.register(Registration.singleton(DisclaimerService, DisclaimerService));
     container.register(Registration.singleton(FirebaseService, FirebaseService));
     container.register(Registration.singleton(FirestoreService, FirestoreService));
-    container.register(Registration.singleton(FirestoreDealsService, FirestoreDealsService));
     container.register(Registration.singleton(EnsService, EnsService));
```
giladt, line 60: What is the reason for removing the `FirestoreDealsService`?

giladt (approved):

src/services/register.ts
IonelLupu: It was already set above using the interface

giladt (approved):

PR #1164, opened 2022-08-11
1098 - daoplomat rewards contract params update

IonelLupu:
## What was done
Updated the algorithm of building the parameters for the deal swap contract so it follows the correct format (I talked with Marvin G about the correct parameters format)

## Testing
Create a deal and follow the steps to execute it. After, check if the primary DAO, partner DAO and daoplomats received the correct token ammount
#### Before
The platform didn't support having the same token address for primary AND partned DAO

#### After
The platform supports the same token address for primary AND partner DAO

### DIFF
--- a/src/entities/DealTokenSwap.ts
+++ b/src/entities/DealTokenSwap.ts
@@ -680,7 +680,10 @@ export class DealTokenSwap implements IDeal {
     const metadata = formatBytes32String(this.id);
     const deadline = 1712882813; // TODO: remove HACK this.fundingPeriod;
 
-    const daoplomatRewards: IDaoplomatRewards = this.registrationData.terms.daoplomatRewards ?? { daoplomats: [], percentage: 0};
+    const daoplomatRewards: IDaoplomatRewards = this.registrationData.terms.daoplomatRewards ?? {
+      daoplomats: [],
+      percentage: 0,
+    };
     const daoplomats = daoplomatRewards.daoplomats.map(daoplomat => daoplomat.address);
     const rewards = [
       [parseInt((daoplomatRewards.percentage * 100).toFixed(0))],
@@ -1027,7 +1030,7 @@ export class DealTokenSwap implements IDeal {
     const pathFrom = tokens.map(tokenAddress => {
       return daos.map(dao => {
         const tokenDetails = dao.tokens.find(details => details.address === tokenAddress);
-        return tokenDetails?.amount ?? 0;
+        return BigNumber.from(tokenDetails?.amount ?? 0);
       });
     });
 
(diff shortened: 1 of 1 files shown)


src/entities/DealTokenSwap.ts
```diff
@@ -1011,41 +1007,42 @@ export class DealTokenSwap implements IDeal {
...
-    const pathTo = new Array<unknown>();
-    const pathFrom = new Array<unknown>();
-    const zero = 0;
-    const fourZeros = [0, 0, 0, 0];
-
-    // eslint-disable-next-line @typescript-eslint/prefer-for-of
-    for (let i = 0; i < this.primaryDao.tokens.length; i++) {
-      if (!tokens.includes(this.primaryDao.tokens[i].address)) {
-        tokens.push(this.primaryDao.tokens[i].address);
-        pathFrom.push([BigNumber.from(this.primaryDao.tokens[i].amount), zero]);
```
hiaux0, line 1025: Why don't we need `BigNumber.from` anymore?

src/entities/DealTokenSwap.ts
```diff
@@ -1011,41 +1007,42 @@ export class DealTokenSwap implements IDeal {
...
-        pathTo.push([
-          ...fourZeros,
```
hiaux0, line 1027: Why these `fourZeros` not needed anymore?

hiaux0 (commented):
Fncality
reward: 4.9%
total: 20%
payout: 0.98 :heavy_check_mark: 

image

src/entities/DealTokenSwap.ts
IonelLupu: Based on Marvin's presentation, pathTo is an array of items that contains of 4 token details per DAO.  Eg when you have 2 DAO (like we usually have) and 2 tokens into play on a deal:
```
pathTo[0] = [  // details for the first token
     100, 0, 330, 10, // details from the FIRST DAO about the first token
     0, 0, 0, 0, // details from the Second DAO about the first token
]
pathTo[1] = [  // details for the second token
     0, 0, 0, 0, // details from the FIRST DAO about the second token
     20, 30, 500, 0, // details from the Second DAO about the second token
]

Example when we have 2 DAOs (like we already have) and 1 token into play (meaning that both primary and partner DAOs swap the same token between them):

```
pathTo[0] = [  // details for the first token
     100, 0, 330, 10, // details from the FIRST DAO  about the first and only token
     20, 30, 500, 0, // details from the Second DAO   about the first and only token
]
```

As you can see, if you have 1 token into play on a deal, the pathTo will contain the details for that token from both DAOs. This fix was added in this PR.

But, in the first example, when you have 2 tokens (or N tokens) in a deal at play, each DAO contributes in a specific portion of the array with 4 values. As you can see, you need to add 4 zeros for the DAO that doesn't set a specific token. `fourZeros` was added just as a helper constant to add those 4 zeros.

hiaux0: > As you can see, you need to add 4 zeros for the DAO that doesn't set a specific token. `fourZeros` was added just as a helper constant to add those 4 zeros. 

Ahhh okay, now with the .map and .flatmap approach, the `0` are added from the nullish coalescing.

hiaux0 (approved):

PR #1165, opened 2022-08-12
1131 - fixed tokens loading

IonelLupu:
## What was done
Increased the waiting time for tokens loading
#### After


hiaux0 (approved):
Appears for me in au2-staging, but also increase wait is okay :man_shrugging:

IonelLupu:
> Appears for me in au2-staging, but also increase wait is okay 🤷‍♂️

Yes. The solution is hacky. The ideal case is not to have those `waitUntil` functions throughout the app. But removing them might need refactoring a bit.

PR #1167, opened 2022-08-15
1149 - moved vesting and cliff period tooltip

IonelLupu:
## What was done
Moved the tooltip based on the requirements in the issue
#### After

### DIFF
--- a/src/resources/elements/formattedNumber/formattedNumber.html
+++ b/src/resources/elements/formattedNumber/formattedNumber.html
@@ -1,6 +1,6 @@
 <template>
   <span
-    ptooltip.bind="!hideTooltip ? value: undefined">
+    ptooltip.bind="hideTooltip ? undefined : value">
     ${text}
   </span>
 </template>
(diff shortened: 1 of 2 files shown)


src/resources/elements/formattedNumber/formattedNumber.html
```diff
@@ -1,6 +1,6 @@
 <template>
   <span
-    ptooltip.bind="value">
+    ptooltip.bind="!hideTooltip ? value: undefined">
```
hiaux0, line 3:
```suggestion
    ptooltip.bind="hideTooltip ? undefined : value">
```

double negation :p

hiaux0 (approved):
fixed

src/resources/elements/formattedNumber/formattedNumber.html
IonelLupu: nice find

hiaux0 (approved):

PR #1168, opened 2022-08-15
1152 - cancel deal wizard redirect

IonelLupu:
## What was done
Redirected to the correct route when canceling a deal wizard
#### After


hiaux0 (approved):
redirected to deal :+1:

PR #1169, opened 2022-08-15
1154 - my deals bug

IonelLupu:
## What was done
Fixed "my deals" button when you have "Partnered deals" tab selected.

First of all, I fixed a bug that showed up when you change the Metamask address. Aurelia, for some reason, cannot handle async router lifecycle hooks. The error was hard to debug but, at the end, I had to remove the `async` keyword from the `attached` lifecycle hook.

This fixed the issue, but in the mean time, I found that the "Open proposal" and "partnered deals" tabs were not displayed if the selected one didn't have any deals. So I fixed that as well. It now shows the other tab if the selected one doesn't have any deals

#### After

### DIFF
--- a/src/deals/list/deals.ts
+++ b/src/deals/list/deals.ts
@@ -174,13 +174,9 @@ export class Deals {
     this.showMine = !this.showMine;
     if (this.showMine) {
       //if showing only "my deals" check to see which tab to display by default if there are no deals in either tab
-      const openDeals = this.isTabVisible(0, this.showMine, this.ethereumService.defaultAccountAddress);
-      const partneredDeals = this.isTabVisible(1, this.showMine, this.ethereumService.defaultAccountAddress);
-      if (openDeals) {
+      const hasPartneredDeals = this.isTabVisible(1, this.showMine, this.ethereumService.defaultAccountAddress);
+      if (!hasPartneredDeals) {
         this.cardIndex = 0;
-      }
-      else if (partneredDeals) {
-        this.cardIndex = 1;
       }
     }
   }
--- a/src/deals/list/deals.ts
+++ b/src/deals/list/deals.ts
@@ -137,7 +137,7 @@ export class Deals {
   }
 
   private getDealsForCardIndex(cardIndex: number, showMine: boolean, address: string): DealTokenSwap[] {
-    if (this.cardIndex === undefined) {
+    if (cardIndex === undefined) {
       return [];
     }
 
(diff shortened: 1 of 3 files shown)


src/deals/list/deals.ts
```diff
@@ -174,14 +174,10 @@ export class Deals {
     this.showMine = !this.showMine;
     if (this.showMine) {
       //if showing only "my deals" check to see which tab to display by default if there are no deals in either tab
-      const openDeals = this.isTabVisible(0, this.showMine, this.ethereumService.defaultAccountAddress);
-      const partneredDeals = this.isTabVisible(1, this.showMine, this.ethereumService.defaultAccountAddress);
-      if (openDeals) {
+      const hasPartneredDeals = this.isTabVisible(1, this.showMine, this.ethereumService.defaultAccountAddress);
+      if (!hasPartneredDeals) {
         this.cardIndex = 0;
       }
```
hiaux0, line 180: Can you give a little background of why the "openDeals" case is not needed anymore? (because default?)

hiaux0 (commented):
fncality -> fixed

src/deals/list/deals.ts
IonelLupu: You are right. I tested more and it seems I have to add the other condition back. I also fixed another old one we had when the account was changed in metamask. (check the removed `async` for that)

hiaux0 (changes requested):
Change request: Looks like you only are filtering for proposal lead?
"a deal is “mine” if I am either the Proposal Lead or a Representative" (https://www.notion.so/primedao/All-Deals-My-Deals-7493545968cb4c0f965e2ec249716310#21393bab719144e5bfa1e5d41931d323)

1. connect as 0xBf3a5599f2f6CE89862d640a248e31F30B7ddF29
2.1 refac/au2 branch -> shows also me as a rep
2.2 your branch -> only prop lead?

IonelLupu:
> Change request: Looks like you only are filtering for proposal lead? "a deal is “mine” if I am either the Proposal Lead or a Representative" (https://www.notion.so/primedao/All-Deals-My-Deals-7493545968cb4c0f965e2ec249716310#21393bab719144e5bfa1e5d41931d323)
> 
> 1. connect as 0xBf3a5599f2f6CE89862d640a248e31F30B7ddF29
>    2.1 refac/au2 branch -> shows also me as a rep
>    2.2 your branch -> only prop lead?

Good find. Fixed

hiaux0 (approved):

PR #1172, opened 2022-08-16
1130 - Cannot select clause to discuss

hiaux0:
## What was done
- always generate id when new clause is added
- extra check for existing clauses, that don't have id (ie. id="")

## Testing
0. Note: following is a private deal. Connect to main testing account.
1. https://prime-deals-dapp-git-fix-1130-cannot-select-cl-b8e086-curvelabs.vercel.app/deal/5fVjBaQc36MxuMGqSzizTG

#### Before
clause is "active"

#### After
can select clause

### DIFF
--- a/src/dealDashboard/dealClauses/dealClauses.html
+++ b/src/dealDashboard/dealClauses/dealClauses.html
@@ -3,8 +3,12 @@
     <h2 class="subtitle heading heading3">Deal Clauses</h2>
   </header>
   <ol>
+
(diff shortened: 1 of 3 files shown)


giladt (approved):
Makes sense. I wonder if we need to keep `addIdsToClauses` in this case, or will it be redundant?

hiaux0:
@giladt 

Good q, as I see it, will still be required for "make an offer" case

PR #1175, opened 2022-08-17
The ptoggle input's `change` event was updated to align with the ptoggle AU2 changes

IonelLupu:
## What was done
The ptoggle input's `change` event was updated to send the value of the toggle button instead of sending an object containing a `checked` property with the value. 

The usage of ptoggle wasn't updated to align with this change.

I updated the usage of ptoggle to align with the AU2 changes we did.

#### After

### DIFF
--- a/src/dealDashboard/dealInfo/dealInfo.html
+++ b/src/dealDashboard/dealInfo/dealInfo.html
@@ -220,10 +220,10 @@
     <div class="infoSection dealPrivacyContainer" if.bind="deal.isPartnered" ref="privateDeal">
       <div class="bodySmallTextBold infoTitle">Private Deal</div>
       <ptoggle
-        disabled.bind="!deal.isAuthenticatedProposalLead"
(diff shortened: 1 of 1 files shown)


src/dealDashboard/dealInfo/dealInfo.html
```diff
@@ -220,10 +220,10 @@ <h1 class="title heading gradient">Deal Information</h1>
     <div class="infoSection dealPrivacyContainer" if.bind="deal.isPartnered" ref="privateDeal">
       <div class="bodySmallTextBold infoTitle">Private Deal</div>
       <ptoggle
-        disabled.bind="!deal.isAuthenticatedProposalLead"
+        disabled.bind="!deal.isAuthenticatedProposalLead || settingPrivacy"
```
IonelLupu, line 223: disabled the toggle button so people won't click on it many times while loading

hiaux0 (approved):
:heavy_check_mark:
- disable while loading
- correct param (thus working for me in preview url)

PR #1176, opened 2022-08-18
1163 - Start funding not redirected to funding page

hiaux0:
## What was done
- increase timeout
- add `?.` check in `DisposableCollection.ts`

#### After
- should redirect 100% of the time to funding page

### DIFF
--- a/src/dealDashboard/dealVotes/dealVotes.ts
+++ b/src/dealDashboard/dealVotes/dealVotes.ts
@@ -9,6 +9,8 @@ import { DiscussionsService } from "../discussionsService";
 import { bindable, DialogDeactivationStatuses, IEventAggregator } from "aurelia";
 import {IRouter} from "@aurelia/router";
 
+const CREATE_SWAP_TIMEOUT = 60000;
+
 export class DealVotes {
   @bindable deal: DealTokenSwap;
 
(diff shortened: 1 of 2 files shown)


src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -101,7 +103,7 @@ export class DealVotes {
     }
 
     if (await this.deal.createSwap()) {
-      await Utils.waitUntilTrue(() => !!this.deal.contractDealId); //have to await this so the contractDealId is populated before redirecting to the funding page
+      await Utils.waitUntilTrue(() => !!this.deal.contractDealId, CREATE_SWAP_TIMEOUT); //have to await this so the contractDealId is populated before redirecting to the funding page
```
IonelLupu, line 106: funded a couple of deals and it seems the error doesn't show up anymore

src/services/DisposableCollection.ts
```diff
@@ -18,7 +18,7 @@ export class DisposableCollection implements IDisposable {
       this._dispose(disposable);
     } else {
       for (disposable of this.disposables) {
-        disposable.dispose();
+        disposable?.dispose();
```
IonelLupu, line 21: why is this needed?

hiaux0: `this.disposables` array had `undefined` as values.
I could have investigate why, but this was the easiest fix

IonelLupu (approved):

PR #1178, opened 2022-08-19
1110 - fixed token status not updating

IonelLupu:
## What was done
fixed token status not updating in real-time when the deal is executed by not displaying sections that need token details before they are loaded

#### Before
<Described in issue>

#### After
<Described in issue>

### DIFF
--- a/src/dealDashboard/dealVotes/dealVotes.ts
+++ b/src/dealDashboard/dealVotes/dealVotes.ts
@@ -17,73 +17,75 @@ export class DealVotes {
   accepting = false;
   declining = false;
 
-  everyTextCopy = [
-    {
-      condition: () => this.deal.isCancelled,
-      voteText: "Deal is cancelled",
-      statusText: "Voting is disabled",
-    },
-    {
-      condition: () => this.deal.isFullyClaimed,
-      voteText: "Deal is completed. All tokens have been claimed!",
-      statusText: "Completed",
-    },
-    {
-      condition: () => this.deal.isClaiming,
-      voteText: "Swap has been executed. Tokens can now be claimed.",
-      statusText: "Claiming",
-    },
-    {
-      condition: () => this.deal.isFunding,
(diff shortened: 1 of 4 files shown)


hiaux0 (approved):

src/dealDashboard/dealVotes/dealVotes.ts
```diff
@@ -17,73 +17,75 @@ export class DealVotes {
...
-      statusText: "Voting is completed",
-    },
-    {
-      condition: () => this.deal.majorityHasVoted && !this.deal.isAuthenticatedRepresentativeUser && this.deal.isAuthenticatedProposalLead,
-      voteText: "You can now initiate the funding phase",
-      statusText: "Voting is completed",
-    },
-  ];
+  get everyTextCopy() {
+    return [
```
hiaux0, line 21: FYI, changing this to getter made `get statusText` trigger on change

src/entities/DealTokenSwap.ts
```diff
@@ -960,7 +960,7 @@ export class DealTokenSwap implements IDeal {
 
       token.claimingClaimed = this.getClaimedAmount(dao, token.address);
       if (tokenClaimableAmounts.size > 0) {
-        token.claimingClaimable = tokenClaimableAmounts.get(token.address);
+        token.claimingClaimable = tokenClaimableAmounts.get(token.address) ?? BigNumber.from(0);
```
hiaux0, line 963: left side could be `undefined`.
I did not dig deeper why, because it seems it should not be allowed to be `undefined` in the first place

IonelLupu: good catch.
_Typescript strict mode would help us catch these bugs early_

src/dealDashboard/dealVotes/dealVotes.ts
IonelLupu: where did yuo change it?

hiaux0: `everyTextCopy` was just a normal class variable before

PR #1181, opened 2022-08-19
1144 - fixed feature of preserving stage settings

IonelLupu:
## What was done
Saved the stage settings in the container so we can access them later if needed. Each stage has its own settings object that can be accessed using the `wizardSettings.<stage route>` injectable string literal

#### Before
This feature was broke by the way AU2 router works so we had to find other solutions. eg: saving settings in the container
#### After
Stage stage is now preserved when switching stages

### DIFF
--- a/src/wizards/tokenSwapDealWizard/wizardManager.ts
+++ b/src/wizards/tokenSwapDealWizard/wizardManager.ts
@@ -249,6 +249,7 @@ export class WizardManager implements IRouteableComponent {
     // It is passed to the wizardService registerWizard method to register it with correct indexOfActive
     this.activeIndex = this.stages.findIndex(stage => stage.route.includes(stageRoute));
 
+    // this is needed to save the stage's state in the current opened wizard
     this.stages.forEach(stage => {
       this.container.register(Registration.instance(`wizardSettings.${stage.route}`, {}));
     });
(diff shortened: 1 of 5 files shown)


src/wizards/tokenSwapDealWizard/wizardManager.ts
```diff
@@ -248,6 +248,10 @@ export class WizardManager implements IRouteableComponent {
     // Getting the index of currently active stage route.
     // It is passed to the wizardService registerWizard method to register it with correct indexOfActive
     this.activeIndex = this.stages.findIndex(stage => stage.route.includes(stageRoute));
+
+    this.stages.forEach(stage => {
+      this.container.register(Registration.instance(`wizardSettings.${stage.route}`, {}));
+    });
```
hiaux0, line 254: Could add a small comment, of what this is doing?

hiaux0 (commented):
## fncality
- confirmed bug on staging
- confirmed fix
- (via clauses)

src/wizards/tokenSwapDealWizard/wizardManager.ts
IonelLupu: added

hiaux0 (approved):

PR #1192, opened 2022-08-23
1185 - removed the error message for the token details you didn't modify

IonelLupu:
## What was done
removed the error message for the token details you didn't modify
#### Before
We used the same method that checks if the tokens are not save for both token details
#### After
Separated the calls for checking if the tokens are saved or not

### DIFF
--- a/src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
+++ b/src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
@@ -19,7 +19,6 @@
         <token-details
           hide-delete-button.bind="primaryDaoTokens.length === 1 && !isOpenProposalWizard"
           on-delete.call="deleteToken($index, primaryDaoTokens, primaryDAOTokenDetails, stageSettings.primaryDAOTokenDetailsViewModes)"
-          on-saved.call="checkedForUnsavedChanges()"
           repeat.for="token of primaryDaoTokens"
           token.bind="token"
           view-mode.bind="stageSettings.primaryDAOTokenDetailsViewModes[$index]"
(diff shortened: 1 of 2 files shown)


src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
```diff
@@ -19,7 +19,6 @@ <h2 class="sidebarHeader heading heading3 title" data-test="section-title">
         <token-details
           hide-delete-button.bind="primaryDaoTokens.length === 1 && !isOpenProposalWizard"
           on-delete.call="deleteToken($index, primaryDaoTokens, primaryDAOTokenDetails, stageSettings.primaryDAOTokenDetailsViewModes)"
-          on-saved.call="checkedForUnsavedChanges()"
```
hiaux0, line 22: Why remove this call altogether?

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
```diff
@@ -56,7 +55,6 @@ <h2 class="sidebarHeader heading heading3 title">Partner DAO Tokens</h2>
         <token-details
           hide-delete-button.bind="partnerDaoTokens.length === 1 && !isOpenProposalWizard"
           on-delete.call="deleteToken($index, partnerDaoTokens, partnerDAOTokenDetails, stageSettings.partnerDAOTokenDetailsViewModes)"
-          on-saved.call="checkedForUnsavedChanges()"
```
hiaux0, line 59: Why remove this call altogether?

hiaux0 (commented):
### fncality
- verified bug
- verified fix :heavy_check_mark:

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
IonelLupu: because this check is already done by the validator

src/wizards/tokenSwapDealWizard/stages/tokenDetailsStage/tokenDetailsStage.html
IonelLupu: because this check is already done by the validator

hiaux0 (approved):

PR #1194, opened 2022-08-24
1191 - prevented going to the submit stage if the other stages are inva…

IonelLupu:
## What was done
prevented going to the submit stage if the other stages are invalid for new deals

#### Before
The terms stage was allowing to proceed if other stages were invalid

### DIFF
--- a/src/wizards/tokenSwapDealWizard/wizardManager.ts
+++ b/src/wizards/tokenSwapDealWizard/wizardManager.ts
@@ -194,9 +194,15 @@ export class WizardManager implements IRouteableComponent {
       return;
     }
 
-    // Make sure all the stages are valid before continuing
-    if (this.wizardType === WizardType.createOpenProposal || this.wizardType === WizardType.createPartneredDeal) {
-      const wizardIsInvalid = this.stages.some(stage => !stage.valid);
+    // Make sure all the stages are valid before continuing to the submit stage
+    const isLastStage = this.activeIndex === this.stages.length - 2;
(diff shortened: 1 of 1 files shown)


src/wizards/tokenSwapDealWizard/wizardManager.ts
```diff
@@ -188,11 +188,21 @@ export class WizardManager implements IRouteableComponent {
...
   public async next() {
     const result = await this.controller.validate();
+    this.stages[this.activeIndex].valid = result.valid;
     if (!result.valid) {
       this.eventAggregator.publish("handleValidationError", "Unable to proceed, please check the page for validation errors");
       return;
     }
 
+    // Make sure all the stages are valid before continuing
+    if (this.wizardType === WizardType.createOpenProposal || this.wizardType === WizardType.createPartneredDeal) {
```
hiaux0, line 198: @IonelLupu 
Do you perhaps know why this has to be added in Au2, but was not needed in Au1?

IonelLupu: it was in AU1 [here](https://github.com/PrimeDAO/prime-deals-dapp/blob/development/src/wizards/services/WizardService.ts#L138)

The wizard was built a bit different but the functionality was almost the same

hiaux0 (approved):
verified bug
verified fix :checkered_flag:

IonelLupu:
@hiaux0  found a bug in this one. I pushed another fix here: https://github.com/PrimeDAO/prime-deals-dapp/pull/1198

PR #1198, opened 2022-08-25
1191 - fixed a bug where it didn't check for visible stages

IonelLupu:
fixed a bug where it didn't check for visible stages but for hidden ones

#### After


hiaux0 (approved):

giladt (approved):
nice one!

PR #1200, opened 2022-08-26
1180- fixed the rounding issues across the app by removing the truncate fun…

IonelLupu:
fixed the rounding issues across the app by removing the truncate function

#### Before
The NumberService was truncating the last digits of a number and it wasn't able to correctly round it to the closest decimal

#### After


hiaux0:
summary:

- rounding can give user inaccurate information
- don't want to round up

Explanation: 
Round down:
- error of precision
Rounding up
- error of precision 
- error accuracy

Example
- 10.5 tokens
- down: 10 is correct information 
- up: 11 wrong information

PR #1208, opened 2022-09-02
get coin id from platforms

testing-life:
## What was done
Another dictionary is created allowing to look for coin id by address, rather than just name 
image

#### After

### DIFF
--- a/src/services/TokenService.ts
+++ b/src/services/TokenService.ts
@@ -107,7 +107,7 @@ export class TokenService {
     return `${name.toLowerCase()}_${symbol.toLowerCase()}`;
   }
 
-  private getTokenGeckoId(name: string, symbol: string, address:string): string {
+  private getTokenGeckoId(name: string, symbol: string, address:string = null): string {
     const id = this.coinAddressIdDictionary.get(address) || this.geckoCoinInfo.get(this.getTokenGeckoMapKey(name, symbol));
     if (id) {
       return id;
@@ -199,7 +199,6 @@ export class TokenService {
           tokenInfo.price = cachedPrice;
         } else {
           if (!tokenInfo.id) {
-            // tokenInfo.id = this.getTokenGeckoId(tokenInfo.name, tokenInfo.symbol);
             tokenInfo.id = this.getTokenGeckoId(tokenInfo.name, tokenInfo.symbol, tokenInfo.address);
           }
           if (tokenInfo.id) {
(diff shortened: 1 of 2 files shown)


blink-km:
@hiaux0 @giladt Could someone review this PR please?

src/services/TokenService.ts
```diff
@@ -71,7 +71,8 @@ export class TokenService {
 
   public async initialize(): Promise<TokenListMap> {
     this.geckoCoinInfo = new Map<string, string>();
-    const uri = `https://pro-api.coingecko.com/api/v3/coins/list?x_cg_pro_api_key=${process.env.COINGECKO_API_KEY}`;
+    this.coinAddressIdDictionary = new Map<string, string>();
```
hiaux0, line 74: There is an `Address` type

src/services/TokenService.ts
```diff
@@ -71,7 +71,8 @@ export class TokenService {
...
+    const uri = `https://pro-api.coingecko.com/api/v3/coins/list?include_platform=true&x_cg_pro_api_key=${process.env.COINGECKO_API_KEY}`;
```
hiaux0, line 75: Would have expected to have some words about `include_platform` in the initial comment pointing to their api docs
https://www.coingecko.com/en/api/documentation

src/services/TokenService.ts
```diff
@@ -80,17 +81,23 @@ export class TokenService {
...
       });
     TimingService.end("get geckoCoinInfo");
-
     return this.tokenLists = await this.tokenListService.fetchLists();
   }
 
   private geckoCoinInfo: Map<string, string>;
+  private coinAddressIdDictionary: Map<string, string>;
 
+  private coinMapping(tokenInfo: ITokenInfo){
```
hiaux0, line 94: return type missing and formatting `){`

src/services/TokenService.ts
```diff
@@ -100,8 +107,8 @@ export class TokenService {
     return `${name.toLowerCase()}_${symbol.toLowerCase()}`;
   }
 
-  private getTokenGeckoId(name: string, symbol: string): string {
-    const id = this.geckoCoinInfo.get(this.getTokenGeckoMapKey(name, symbol));
+  private getTokenGeckoId(name: string, symbol: string, address:string = null): string {
```
hiaux0, line 110: with just type `:string` I would expect the default to be `""` and not `null`

src/services/TokenService.ts
hiaux0: But okay, the above Map also just has `string`

tamagosante (approved):
### Fncality 
:heavy_check_mark: 
image


### Code

